Issue #1, opened 2020-06-29
Support GeoJSON literals in GeoSPARQL


dr-shorthair:
New datatype `rdf:JSON` is introduced in [JSON-LD Syntax](https://w3c.github.io/json-ld-syntax/#the-rdf-json-datatype)

See https://github.com/w3c/json-ld-syntax/issues/353

jabhay:
MoSCoW poll created

[![](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Must)](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Must/vote)
[![](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Should)](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Should/vote)
[![](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Could)](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Could/vote)
[![](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Won%27t)](https://api.gh-polls.com/poll/01EJXBPCEDYBJ7EWE3X67Q17YT/Won%27t/vote)

dr-shorthair:
Discussion of this topic has moved to #48

dr-shorthair:
Whoops - #48 is matching PR.

mathib:
Is anyone aware of the apparent overlap with [this](https://www.ogc.org/standards/eo-geojson) OGC GeoJSON(-LD) extension specification? 

They mention GeoSPARQL with WKT literals as an alternative for GeoJSON-LD, since at the time of writing, the JSON-LD 1.0 spec did not support lists of lists (the current 1.1 version does btw).

In the revision of GeoSPARQL, we now defined an alternative approach based on GeoJSON literal, so there's no actual overlap with the above OGC standard, but there might be room for (initial) confusion. I think we should at least mention this OGC spec in the revisioned GeoSPARQL spec? Maybe the contributors to the OGC GeoJSON(-LD) extension specification can also provide input for our GeoSPARQL revision?

situx:
[...]

Issue #42, opened 2020-10-03
separate geometry properties for coordinates and type


FransKnibbe:
This issue corresponds with [OGC Standards tracker request 659](http://ogc.standardstracker.org/show_request.cgi?id=659)

A geometry can be considered as consisting of different elements (properties), at least the following:
1) Coordinate Reference System used
2) Geometry type (line, polygon, multipolygon, ...)
3) An array of coordinates
4) Dimension (0, 1, 2 or 3)
5) Spatial resolution (level of detail)

None of these properties should be mandatory in a description of a geometry.

Addings some of these properties is already covered by other change requests for GeoSPARQL. That is why this request is for adding extra properties to the geometry class for geometry type and the coordinate array. Having them available as separate properties will allow for publication of data where one of the properties is unknown, and will improve data selection and reasoning capabilities of data stores.

situx:
I think there could be a dimension value of 4 if we include measurement coordinates which are defined in WKT (https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry) (and I believe also in WKT literals?)

FransKnibbe:
@situx: thank you for drawing attention to linear referencing. True, in WKT the measurement value (M) acts as a possible fourth or third coordinate. But in terms of a spatial ontology my feeling is that it seems best to treat measurements on a spatial object as something separate from the spatial shape. I don't think any particular data serialization (such as WKT) should influence the core ontology.
It seems [issue 21](https://github.com/opengeospatial/ogc-geosparql/issues/21) is a good starting point for discussing how linear referencing should be modelled exactly.

nicholascar:
> ..extra properties to the geometry class for geometry type and the coordinate array...

The ontology already defines a property for coordinate array: `geo:hasSerialization`. This is the super property of the more commonly used `geo:asWKT` and `geo:asGML`. 

The property definition reads "Connects a geometry object with its text-based serialization." and this is better than saying _coordinates_ as some geometries done use coordinates but ordinates (e.g. DGGS lists of Cell IDs) however all are amenable to "text-based serialization".

I will suggest elsewhere that we re-focus on this property and avoid declaring a property per serialization type like `geo:asWKT` so that we can instead focus on a vocab of "text-based serializations" data types to be published as another profile artefact by the Naming Authority alongside the ontology, but one that can be grown independently of the ontology revision cycle.

FransKnibbe:
@nicholascar: good to point out that ordinates could also be used. 
[...]

Issue #31, opened 2019-10-18
Decouple CRS and WKT


lvdbrink:
**Official change request**: [OGC CR 595](http://ogc.standardstracker.org/show_request.cgi?id=595)

WKT seems a good way to easily encode geographic geometry, but the datatype geo:wktLiteral makes it hard to work with GeoSPARQL. In a next version, there should be an option to use only a WKT literal and use a different way to express the CRS of a geometry. Reasons why concatenation of CRS URI and WKT can be considered bad design are:
- GeoSPARQL deviates from the WKT standard, resulting in poor software support.
- Allowing not to specify a CRS and defaulting to CRS84 may be useful in North America, but is of little value for serious usage in other parts of the world.
- The proper data type for expression of CRS is an IRI. Therefore it should be defined as such, not as part of a string literal.

Especially when non-geographical geometry is considered, CRS is not necessarily a known property. Therefore it should be possible to leave out CRS data in publications, without this resulting in wrong interpretations.

CRS can be considered an intrinsic or fundamental aspect of geometry, but so are other properties like dimensionality or accuracy. This does not mean all of this information should be lumped together in one literal.

It seems better to introduce a new property for CRS and to let WKT literals be just WKT literals. Should a new property for indicating CRS be introduced, it would be good to allow it to be applied not only to individual geometries, but to geometry collections too.

jabhay:
Created MoSCoW poll

[![](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Must)](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Must/vote)
[![](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Should)](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Should/vote)
[![](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Could)](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Could/vote)
[![](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Won%27t)](https://api.gh-polls.com/poll/01EJS8Z01Y4AF0B4VTQV56CWSN/Won%27t/vote)

mperry455:
I have a few concerns with this decoupling.

[...]

PR #48, opened 2020-10-30
GeoJSON literals for GeoSPARQL

situx:
I have extended the Geometry Extension section of the GeoSPARQL specification to include a definition for a GeoJSON literal.
The ontology has also been extended by a geo:asGeoJSON property and a geo:geoJSONLiteral.

I had the following considerations:
- I chose to represent GeoJSON using a GeoJSON literal and not using the rdf:json relation because other serializations of geometries in JSON exist, for example ESRIJSON, TopoJSON, or even pure JSON files with lat/lon geocoordinates attached. Whether or not to leave it this way is a matter of discussion
- GeoJSON literals do not allow the definition of a CRS in the literal as opposed to WKT and GML as the GeoJSON standard only allows one CRS (WGS84) to be used
- GeoJSON literals are defined by the Geometry Object of the GeoJSON specification, e.g. {"type":"Point","coordinates":[-83.38,33.95]}
- An empty literal is equivalent to the statement {"geometry":null} 
However, how do we define equivalences for POINT(EMPTY), POLYGON(EMPTY) etc.? Is this necessary?
Should we define an empty Point as {"type":"Point","coordinates":[]} a.s.o. ?
- Requirement definitions currently do not contain a number as we might modify the numbering later on

I was not sure whether it is needed to extend the RDFS Entailment section.
In my opinion, we could just reuse the Geometry class definitions of SimpleFeatures which were used for defining the WKTLiterals.
But I am open to any other suggestions.

### DIFF
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -181,7 +181,6 @@ skos:prefLabel a owl:AnnotationProperty .
 	rdfs:comment """
       A GeoJSON serialization of a geometry object.
     """@en ;
-	rdfs:subClassOf rdf:JSON ;
 	rdfs:seeAlso <https://tools.ietf.org/html/rfc7946> ;
 	rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.1> ;
 	rdfs:label "GeoJSON Literal"@en ;
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -167,6 +167,27 @@ skos:prefLabel a owl:AnnotationProperty .
       A Well-known Text serialization of a geometry object.
     """@en ;
 	skos:prefLabel "Well-known Text Literal"@en .
+	
+# 
+# http://www.opengis.net/ont/geosparql#wktLiteral
+
+:geoJSONLiteral a rdfs:Datatype ;
+	dc:contributor "Timo Homburg" ;
+	dc:creator "OGC GeoSPARQL 2.0 Standard Working Group" ;
+	dc:date "2020-10-30"^^xsd:date ;
+	dc:description """
+      A GeoJSON serialization of a geometry object.
+    """@en ;
+	rdfs:comment """
+      A GeoJSON serialization of a geometry object.
+    """@en ;
+	rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.1> ;
+	rdfs:label "GeoJSON Literal"@en ;
+	skos:definition """
+      A GeoJSON serialization of a geometry object.
+    """@en ;
+	skos:prefLabel "GeoJSON Literal"@en .
+	
 # 
 # http://www.w3.org/2001/XMLSchema#date
 
@@ -864,6 +885,30 @@ xsd:date a rdfs:Datatype .
       The WKT serialization of a geometry
     """@en ;
 	skos:prefLabel "asWKT"@en .
+	
+# 
+# http://www.opengis.net/ont/geosparql#asGeoJSON
+
+:asGeoJSON a owl:DatatypeProperty ;
+	rdfs:subPropertyOf :hasSerialization ;
+	rdfs:domain :Geometry ;
+	rdfs:range :geoJSONLiteral ;
+	dc:contributor "Timo Homburg" ;
+	dc:creator "OGC GeoSPARQL 2.0 Standard Working Group" ;
+	dc:date "2020-10-30"^^xsd:date ;
+	dc:description """
+      The GeoJSON serialization of a geometry
+    """@en ;
+	rdfs:comment """
+      The GeoJSON serialization of a geometry
+    """@en ;
+	rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.1> ;
+	rdfs:label "asGeoJSON"@en ;
+	skos:definition """
+      The GeoJSON serialization of a geometry
+    """@en ;
+	skos:prefLabel "asGeoJSON"@en .	
+
 # 
 # http://www.opengis.net/ont/geosparql#coordinateDimension
 
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -181,6 +181,7 @@ skos:prefLabel a owl:AnnotationProperty .
 	rdfs:comment """
       A GeoJSON serialization of a geometry object.
     """@en ;
+	rdfs:seeAlso <https://tools.ietf.org/html/rfc7946> .
 	rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.1> ;
 	rdfs:label "GeoJSON Literal"@en ;
 	skos:definition """
@@ -893,6 +894,7 @@ xsd:date a rdfs:Datatype .
 	rdfs:subPropertyOf :hasSerialization, rdf:json ;
 	rdfs:domain :Geometry ;
 	rdfs:range :geoJSONLiteral ;
+	rdfs:seeAlso <https://tools.ietf.org/html/rfc7946> .
 	dc:contributor "Timo Homburg" ;
 	dc:creator "OGC GeoSPARQL 2.0 Standard Working Group" ;
 	dc:date "2020-10-30"^^xsd:date ;
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -181,7 +181,8 @@ skos:prefLabel a owl:AnnotationProperty .
 	rdfs:comment """
       A GeoJSON serialization of a geometry object.
     """@en ;
-	rdfs:seeAlso <https://tools.ietf.org/html/rfc7946> .
+	rdfs:subClassOf rdf:JSON ;
+	rdfs:seeAlso <https://tools.ietf.org/html/rfc7946> ;
 	rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.1> ;
 	rdfs:label "GeoJSON Literal"@en ;
 	skos:definition """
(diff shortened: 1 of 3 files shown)


dr-shorthair:
Also see #1

FransKnibbe:
It seems funny to me that the GeoJSON serialisation is only applicable to geometries that happen to use CRS84. OK, it is mentioned in the documentation. But the ontology itself does not prohibit erroneous usage. 
One solution I can think of is to add a subclass of geometry that has CRS84 as a fixed CRS, e.g. geo:Crs84Geometry, and make the serialisation applicable to only that class.

Secondly, shouldn't there be a reference to the GeoJSON specification that should be used (ideally a URI) and the exact version of that specification?

situx:
Hi Frans,

the reference given in the specification is [RFC 7946] https://tools.ietf.org/html/rfc7946
Do you mean to add this reference to the ontology specification?
Whether to add a new class which enforces CRS84, I would leave that open for discussion. 
I currently cannot think of an optimal solution.

FransKnibbe:
Hi Timo,

Yes, I think it is a good idea to add a reference to the applicable specification to the ontology. Perhaps rdfs:seeAlso is an appropriate property to use if we want to do that?

nicholascar:
1. please could the proposed datatype `geo:geoJSONLiteral` be a subPropertyOf the new `rdf:JSON` [datatype, as proposed in JSON-LD 1.1](https://www.w3.org/TR/json-ld11/#the-rdf-json-datatype)?

2. I wonder about all this WGS84 stuff. I'm no fan of locking in a single CRS and thus I have problems with GeoJSON. Perhaps we should provide both a JSON format with CRS unspecified and GeoJSON that is CRS locked. If we don't indicate how to generalise these things in GeoSPARQL, who else will?

dr-shorthair:
> It seems funny to me that the GeoJSON serialisation is only applicable to geometries that happen to use CRS84. OK, it is mentioned in the documentation. 

I'm afraid that is a feature of GeoJSON. 

There was proposal to allow CRS to be indicated in GeoJSON , but the developers were strongly of the opinion that it was best to limit it to the common case seen in most web-mapping applications and platforms, which were exploding in popularity at the time GeoJSON was designed. It was targeted at a distinct market, not professional geographers or cartographers or geodesists, that was judged to be unable and unwilling to go into the details of CRS. It is all history now - GeoJSON is what it is. 

Other JSON encodings might be developed which don't have those limitations, potentially as supersets of GeoJSON, but those should not be called 'GeoJSON'.

dr-shorthair:
Fixes #1 
(when done)

nicholascar:
> Other JSON encodings might be developed which don't have those limitations, potentially as supersets of GeoJSON, but those should not be called 'GeoJSON'.

Yes, what to do for a GeoJSON-like encoding of data within OGC APIs that use CRSes other than WGS84? I've used a GeoJSON-like format in my DGGs APIs and made up a "crs" property...

FransKnibbe:
Frankly, I am not a fan of GeoJSON either. I have a hard time imagining why anyone would want to use it. Sure, JSON is a very handy format to use on web pages. But JSON-LD seems much more useful. I think JSON-LD could be even more useful for use with spatial data if GeoSPARQL were to have a proper full geometry model ([see issue 42](https://github.com/opengeospatial/ogc-geosparql/issues/42))

As it stands, I don´t  think we can speak of a GeoJSON serialisation of a GeoSPARQL geometry. It can only serialize a subset of geometries. Hence the suggestion to actually define that subset. But in order to do that it would be good to have a separate CRS property for geometry ([see issue 31](https://github.com/opengeospatial/ogc-geosparql/issues/31)).

nicholascar:
> I don´t think we can speak of a GeoJSON serialisation of a GeoSPARQL geometry.

I think we can, as long as there's a description of how the data is "downscaled" into the lesser content format of GeoJSON. So, the representation a. only works for WGS84 geometries and b. looses semantics of properties so that any properties defined in RDF just become key/value pairs in the GeoJSON with no property definitions, no explicit datatypes etc.

I think we _should_ provide this downscaling mapping at the very least to indicate what you loose when you use GeoJSON.

In my OGC API implementations, I apparently _must_ use GeoJSON but I have semantic content in the DB so I have to do this downscaling!

FransKnibbe:
@nicholascar: good to point out the possible loss of information when RDF data are exported to GeoJSON. If GeoJSON support is added to GeoSPARQL, the reverse should also be well thought out: What happens when GeoJSON literals are imported into an RDF storage system / graph store? GeoSPARQL implementers will have to try to make sense of geometry properties in GeoJSON, and loss of information can be incurred in that direction too.

It seems to me that GeoJSON support could place a heavy burden on implementers, with a danger of different implementers choosing different solutions. This in turn makes me wonder if we can't place GeoJSON support in a separate module, so that supporting it becomes more optional. 

I think my main concern with support for GeoJSON literals is that it might impede further development (extension) of the geometry class in GeoSPARQL.

nicholascar:
> What happens when GeoJSON literals are imported into an RDF storage system / graph store?

Well there are currently no _GeoSPARQL_ rules about how to handle any of the GeoSPARQL literals (WKT or GML). There are no ontology conversion functions specified in _ontology_ code. Any functions doing things with the literals, like spatial indexing, are implementation-specific i.e. StarDog perhaps performs some sort of GDAL indexing of WKT & GML literals. So I don't think we - ontology designers - have to worry about this. We just state that GeoJSON literals are a microformat in GeoSPARQL and any handling is implementation specific.

If someone wanted to mapp GeoJSON literals to semantic content then that would be a matter for application node, not GeoSPARQL ontology rules.

I will be doing something similar with DGGS literals: the ontology doesn't know anything about the DGGS literal content: handling those is also a job for application code, even if they have GeoSPARQL functions specified for them.

--

We have plenty of lossy literals in Semantic Web for precedence here. Essentially any microformat that isn't understood by an ontology. This would include custom identifier properties that contain structured information as recommended for use in SKOS. So SKOS doesn't know anything about the internals of the identifier literals...

situx:
I have a question especially concerning OGC API Features.
In the CRS Extension of OGC API Features (http://docs.opengeospatial.org/DRAFTS/18-058.html#crs-query) we face a similar problem. 
It is stated in section 6.3.3.2. Features and Feature resources:
"GeoJSON normatively supports WGS 84 (without height: CRS84; with height: CRS84h), but the "prior arrangement" provision allows other coordinate systems to be used."
What is this prior arrangement provision and may we do a similar approach here?

nicholascar (approved):
Approved for this to be added to the ontology as long as this and other literal types can also be added to a register of literals

situx:
I changed the RDFS Entailment section to use classes of the GeoJSON vocabulary which is used in GeoJSON-LD (https://geojson.org/geojson-ld/)
However, this vocabulary does not define a Geometry class, only classes for {Multi}Point,{Multi}LineString,{Multi}Polygon.
I would suggest to use the GeoJSON vocabulary but to set all Geometry classes of the GeoJSON vocabulary as subclasses of sf:Geometry.
I am not sure if this is a perfect solution, but it might do.

situx:
After today's discussion I removed the RDFS entailment section part and checked the empty geometry literal which is now defined only as {"geometry":null}.
In my view the change request is therefore ready to be merged and I am looking forward to a second reviewer.

dr-shorthair:
Should it also be ? 
```
:geoJSONLiteral rdfs:subClassOf rdf:JSON .
```

situx:
Thanks @dr-shorthair it is now corrected

mperry455:
Are we sure that :geoJSONLiteral should be subclass of rdf:JSON?
rdf:JSON has some rules about canonical representation (no extra whitespace, keys ordered lexicographically, etc.)
https://www.w3.org/TR/json-ld/#the-rdf-json-datatype
I don't know if we want to require geoJSON strings to match these rules.

situx:
Good point. Even though the literal would just include two keys, in my opinion key ordering is not necessary. But I wonder how others think about it?

mperry455 (changes requested):
I suggest that we remove rdfs:subClassOf rdf:JSON to avoid the canonicalization requirements of rdf:JSON. gmlLiteral is not a subClassOf rdf:XMLLiteral for this same reason.

situx:
I removed the subclass relation for now. If this needs further discussion then we may still do so tomorrow

mperry455 (approved):
Thanks Timo. Looks good.

nicholascar:
Good pickups Simon & Matt: I had assumed we _would_ want `:geoJSONLiteral rdfs:subClassOf rdf:JSON .` but if you foresee issues there, we don't have to. I guess the spirit of GeoSPARQL is that information about those literal microformats is elsewhere so users will just have to know all about GeoJSON (and not necisarily rdf:JSON) as discussed in the last teleconference.

andrea-perego:
This is just to report that the adoption of this newly defined datatype is being considered in [GeoDCAT-AP](https://semiceu.github.io/GeoDCAT-AP/drafts/latest/) (see https://github.com/SEMICeu/GeoDCAT-AP/issues/4). 

Actually, the only concern we have is that the GeoJSON datatype may be dropped or changed in the final version of GeoSPARQL - we are going to release the new version of GeoDCAT-AP shortly, so we cannot wait until then.

About the context:

GeoDCAT-AP is a profile of the W3C DCAT vocabulary used across Europe to support a DCAT-compliant representation of INSPIRE / ISO 19115 metadata. Some more details are also available in the GeoDCAT-AP OGC Discussion Paper: http://www.opengis.net/doc/dp/GeoDCAT-AP

nicholascar:
> the only concern we have is that the GeoJSON datatype may be dropped or changed in the final version of GeoSPARQL

It won't be. Everyone wants it included so the only issues are around the datatype formalisms like the conversation above re alignment with `rdf:JSON` etc. So use it with confidence

mathib:
note/question: is it legal to define subdatatypes in domain ontologies used for reasoning? I always imagined that it is not allowed as there's no `rdfs:subDatatypeOf` relation... I used the [OWL profilechecker](https://github.com/stain/profilechecker) which renders errors when using `rdfs:subClassOf` on datatypes "_Datatype IRI also used as Class IRI_"). My guess is that it's theoretically allowed (e.g. as done in the [RDF file](https://www.w3.org/1999/02/22-rdf-syntax-ns#) describing RDF terminology for `rdf:JSON rdfs:subClassOf rdfs:Literal`), but not encouraged in domain ontologies as that might confuse reasoning engines (similar as users are not encouraged to directly feed the RDF files describing RDFS, RDF, OWL, etc. to a reasoning engine).

Issue #7, opened 2020-01-26
Add support for Spatial Aggregate functions


situx:
Geospatial literals currently cannot be queried using traditional aggregate functions such as MIN,MAX,AVG.
However it may be useful to get the minimum or maximum X,Y or Z coordinate for a given set of geometries which is currently not possible in GeoSPARQL.
One use case could be: Finding a bounding box of a set of geometries for which a minimum X,Y and maximum X,Y coordinate of all geometries within a graph would be needed.
To my knowledge there is no such functionality currently available in GeoSPARQL.
My proposal:
Adding the following aggregate functions:
MINX, MINY, MAXX, MAXY, MINZ, MAXZ
In addition adding a BBOX aggregate method which would calculate a minimum bounding box of a set of geometries bound to a query variable.
I have not yet found a use case for an AVG aggregate function so I would neglect it, however, maybe there are some ideas about that from the community.

jabhay:
MoSCoW poll created

[![](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Must)](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Must/vote)
[![](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Should)](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Should/vote)
[![](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Could)](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Could/vote)
[![](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Won%27t)](https://api.gh-polls.com/poll/01EJXBGJ6XN1C93T85X2FFJ8HJ/Won%27t/vote)

situx:
Even though we marked this issues as GeoSPARQL 1.1, we should think about which aggregates to define for GeoSPARQL 1.1.
I would include Aggregate functions for a Z coordinate, as the literals which are currently supported in GeoSPARQL all support Z coordinates.
If in the standardization process we decide to also accept T and L coordinates, then we should also add Spatial aggregate functions accordingly (MINT, MAXT etc.)
I am a bit undecided about M coordinates.
[...]

PR #57, opened 2020-11-19
SpatialAggregate functions

situx:
This is a proposal to include Spatial Aggregate functions in GeoSPARQL. 
The proposal closes issue #7 once approved.
The functions proposed were discussed in issue #7 but additional functions if required may be discussed in this pull request.
I added function signatures according to how non-topological functions were described in the standard before.
I would appreciate feedback on the following issues:
- I am not sure about the syntax to describe a set/list of geometries, so I would appreciate some help if what I did is correct.
- The defined Spatial Aggregate Functions would need to be added to the GeoSPARQL functions vocabulary which has not yet been committed to the repository. So this part is still missing in this pull request.
- Question: Should we add a minimum bounding circle function as well as we have a minimum bounding box function?

Looking forward to your feedback.


1.1/spec/11-Part-08.adoc
```diff
@@ -481,3 +481,100 @@ geof:getSRID (geom: ogc:geomLiteral): xsd:anyURI
...
+geof:minZ (geom: ogc:geomLiteral): xsd:double
+```
+
+Returns the minimum Z coordinate for `geom`.
+
+=== Spatial Aggregate Extension
+
+This clause establishes the Spatial Aggregate extension requirements class with URI `/req/spatial-aggregate-extension`.
+Spatial Aggregate Functions may be used in the SELECT clause of a SPARQL query and in the WHERE clause.
+
```
mperry455, line 536: I would suggest adding one more sentence. Something like:
Such functions calculate an aggregate value over a group of solutions. Solution groups are determined by a GROUP BY clause. All solutions form a single group if no GROUP BY is specified.

With this extra wording, I think we could change Set<ogc:geomLiteral> to ogc:geomLiteral.

1.1/spec/11-Part-08.adoc
```diff
@@ -481,3 +481,100 @@ geof:getSRID (geom: ogc:geomLiteral): xsd:anyURI
...
+```
+geof:minZ (geom: ogc:geomLiteral): xsd:double
+```
+
+Returns the minimum Z coordinate for `geom`.
+
+=== Spatial Aggregate Extension
+
+This clause establishes the Spatial Aggregate extension requirements class with URI `/req/spatial-aggregate-extension`.
+Spatial Aggregate Functions may be used in the SELECT clause of a SPARQL query and in the WHERE clause.
```
mperry455, line 535: I'm having a hard time seeing where aggregates can be used in a WHERE clause except in nested subqueries. Should we say "used in the SELECT, HAVING and ORDER BY clauses of a SPARQL query"

1.1/spec/11-Part-08.adoc
situx: Thanks, I added your suggestion to the draft

1.1/spec/11-Part-08.adoc
situx: Sorry, my bad. The WHERE was a remnant of when I first wrote this paragraph with also the minX etc. functions in mind.
I think SELECT, HAVING and ORDER BY seems reasonable to me. 
I added it to the draft.

nicholascar:
@situx you'll need to add the functions to the ontology RDF document as well as the vocabularies of functions!

I'll add in the vocabulary/vocabularies of functions as soon as the _Profile_ for 1.0 is complete as it contains the files and I'll just be copying it/them across from there in to the 1.1 folder.

situx:
@nicholascar I added the functions, but was unsure if aggregate functions would need a separate namespace. At least the convexHull function is duplicated for now.

mperry455:
I would be in favor of a separate namespace, especially since we have a conflict with convexHull.

mperry455 (approved):

1.1/spec/11-Part-08.adoc
```diff
@@ -481,3 +481,109 @@ geof:getSRID (geom: ogc:geomLiteral): xsd:anyURI
...
+
+```
+geosaf:Centroid (ogc:geomLiteral): ogc:geomLiteral
+```
+Calculates the centroid of the set of given geometries.
+
+===== Function: geosaf:Concat_Lines
+
+```
+geosaf:Concat_Lines (ogc:geomLiteral): ogc:geomLiteral
```
nicholascar, line 565: Can you please change this function name from `Concat_Lines` to `ConcatLines` for consistent styling?

1.1/functions.ttl
```diff
@@ -0,0 +1,27 @@
+PREFIX safuncs: <http://www.opengis.net/def/function/aggregate/geosparql/>
+PREFIX funcs: <http://www.opengis.net/def/function/geosparql/>
+
+funcs: a skos:Collection ;
```
nicholascar, line 4: You need a ConceptScheme & Concept declarations as well as the Collections with `skos:member` elements so that this vocabulary can be properly picked up by the vocab publication tools (not just pyLODE but also the OGC's vocab server).

1.1/functions.ttl
```diff
@@ -0,0 +1,27 @@
+PREFIX safuncs: <http://www.opengis.net/def/function/aggregate/geosparql/>
```
nicholascar, line 1: Can this file be renamed from `functions.ttl` to `sa_functions.ttl` to make it more explicit?

Also, you'll need to add this resource to the Profile declaration, just copy this one:

* <https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/profile.ttl#L84>

nicholascar:
Hi @situx no problems with the content of all of this, just need to ensure that the functions list is listed in the profile and that the vocbaulary is properly constituted with `ConceptSchem` etc, as per the comments above.

nicholascar (changes requested):
As per my other 3 comments (I have to add this comment in to make GH work)

situx:
@nicholascar I think I incorporated all changes that you requested.

situx:
I also updated the CI process so it should build documentation for the spatial aggregate functions

nicholascar (approved):

Issue #56, opened 2020-11-18
GeoSPARQL source code for functions and rules is missing 


FransKnibbe:
The GeoSPARQL semantic code consists of three sets of definitions: the core ontology, the functions vocabulary and the RIF rules. These correspond with the following name spaces:
geo: http://www.opengis.net/ont/geosparql#
geof: http://www.opengis.net/def/function/geosparql/
geor: http://www.opengis.net/def/rule/geosparql/

We will want to keep the three files mutually consistent, and there are change requests for GeoSPARQL functions. So I suggest adding the functions and rules files to [this repository](https://github.com/opengeospatial/ogc-geosparql/tree/master/1.1).

A code file for the functions is mentioned in the OGC definition server: http://defs-dev.opengis.net/static/definitions/conceptschemes/functions_geosparql.ttl. I could not find a source file for the RIF rules, but it must exist somewhere.

FransKnibbe:
The functions file has been found: [http://defs-dev.opengis.net/static/definitions/conceptschemes/functions_geosparql.ttl
](http://defs-dev.opengis.net/static/definitions/conceptschemes/functions_geosparql.ttl)

jabhay:
Rules here perhaps: http://defs.opengis.net/elda-common/ogc-def/resource?uri=http://www.opengis.net/def/rule/geosparql/&_format=html

dr-shorthair:
@rob-metalinkage or @ghobona might be able to advise on status and point-of-truth for these

nicholascar:
@dr-shorthair we have copies of the resources and links to their existing RDF locations in a profile declaration for GeoSPARQL 1.0 that is currently a PR: #58 (see profile.ttl: https://github.com/opengeospatial/ogc-geosparql/pull/58/files#diff-8b9fe01200dc518bf97ace4e2a27c8665a486c8b1000f9d2054257b01ec804fc)

ghobona:
The ontology approved with GeoSPARQL 1.0 is at http://schemas.opengis.net/geosparql/1.0/geosparql_vocab_all.rdf

The functions are listed in there.

jabhay:
Hi @FransKnibbe , could you please review the pull request that I've tagged you in as it should address this issue. Any further questions, please raise comments in the pull request.

nicholascar:
All GeoSPARQL 1.0 and 1.1 resources are listed in this repository at:

[...]

PR #58, opened 2020-11-20
1.0 profile

### LINKED ISSUE #55: Publish GeoSPARQL 1.0 as a profile
Currently GeoSPARQL 1.0 is presented as an ontology and a specification. The ontology is available officially via its namespace URI, <http://www.opengis.net/ont/geosparql>, as an RDF/XML artefact only. The specification is a PDF, officially available at its document URI <http://www.opengis.net/doc/IS/geosparql/1.0>.

This proposal is to present GeoSPARQL 1.0, i.e. the current GeoSPARQL, unchanged, as a _profile_ according to the [Profiles Vocabulary](https://w3c.github.io/dx-prof/prof/). A profile here is a potentially multi-part standard with parts playing roles. Parts may be the PDF specification (role: "specification"), the ontology in RDF form (role:scheme), the ontology in HTML form (role:guidance) etc. The Profiles Vocabulary was established as a W3C vocabulary by OGC staff for this task and is now in use within the OGC Naming Authority.

I propose the following profile definition be submitted to the OGC NA along with the current GeoSPARQL 1.0 ontology RDF and GeoSPARQL 1.0 specification document for their publication via the NA's Name Server.

Profile definition:

```turtle
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX prof: <http://www.w3.org/ns/dx/prof/>
PREFIX role: <http://www.w3.org/ns/dx/prof/role/>
PREFIX sdo: <https://schema.org/>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>


<http://www.opengis.net/def/geosparql>  # not the use of the /def/ as opposed to the ont's /ont/
    a prof:Profile ;
    dcterms:identifier "geosparql" ;
    dcterms:created "2009-10-27"^^xsd:date ;
    dcterms:modified "2012-07-19"^^xsd:date ;
    dcterms:issued "2020-xx-xx"^^xsd:date ;
    dcterms:publisher <https://www.ogc.org> ;
    dcterms:title "GeoSPARQL Specification"@en ;
    dcterms:description "This profile supports representing and querying geospatial data on the Semantic Web. 
                        GeoSPARQL defines a vocabulary for representing geospatial data in RDF, and it 
                        defines an extension to the SPARQL query language for processing geospatial data. It 
                        is presented in a series of artefacts: the vocabulary as an RDF ontology and the 
                        specification as a PDF document." ;
    prof:hasResource 
         <http://www.opengis.net/doc/IS/geosparql/1.0> ,
         <http://www.opengis.net/ont/geosparql> ;
.

<https://www.ogc.org>
        a owl:NamedIndividual ;
        a sdo:Organization ;
        sdo:name "Open Geospatial Consortium" ;
        sdo:url <https://www.ogc.org> ;
.

<http://www.opengis.net/doc/IS/geosparql/1.0>
        dcterms:title "Ontology specification in PDF"@en ;
        # dcterms:conformsTo <xxxx> ; # check for OGC conformance classes for a spec
        dcterms:format "application/pdf" ;
        prof:hasArtifact <https://portal.ogc.org/files/?artifact_id=47664> ;
        prof:hasRole role:specification ;
.

<http://www.opengis.net/ont/geosparql>
        dcterms:title "Ontology document in RDF"@en ;
        dcterms:conformsTo <http://www.w3.org/TR/owl2-rdf-based-semantics/> ;
        dcterms:format "application/rdf+xml" ;
        prof:hasArtifact <http://schemas.opengis.net/geosparql/1.0/geosparql_vocab_all.rdf> ;
        prof:hasRole role:scheme ;
.
```

Once GeoSPARQL 1.0 has been established as a _profile_, GeoSPARQL 1.1 can be worked on as a profile. This will allow for multiple parts of the specification to be related nicely.

nicholascar:
A profile definition for the 1.0 specification with the Spec (PDF doc) and the Ont unchanged. The Funcs/Rules vocab is added after a tidy-up of its RDF (no functional changes).

Closes #55 

@ghobona I've added you as a reviewer for this form of profile definition is what I expect this, and other SWGs (e.g. DGGS) to submit to the NA for profile publication

Notes to reviewers: 

- the form of the profile here is expected, as per the Profiles Vocabulary, but some of the content is necessarily new, in articular the introduction of the profile URI of [http://www.opengis.net/**def**/geosparql](http://www.opengis.net/def/geosparql) as distinct from the Spec document or Ont URIs (<http://www.opengis.net/doc/IS/geosparql/1.0> & <http://www.opengis.net/ont/geosparql> respectively)
- I've made no changes to the spec doc or the ont but some tidy-up of the ont doc (non-functional) may be proposed separately, inline with RDF neatness (removal of accidentally included junk, spaces etc, as already applied to the 1.1 ont)

- the vocab introduced is a tidied version of the "Functions" vocab already auto-generated by the NA but this version has many improvements, such as better naming (in includes both Functions and Rules), definition of a ConcpetScheme as well as the 2 x Collections etc. No changes to the Functions & Rules themselves.


situx (approved):
In general I approve these changes, but would like to ask if it is necessary to also include the GML geometry hierarchy or at least a reference to it in the profile.ttl

jabhay:
Changing approver to @FransKnibbe , owing to this pull request covering off issue #56

FransKnibbe:
Do want to see this PR as a definitive profile for GeoSPARQL 1.0? Or as the start of it, to be open to possible changes later on?
I am asking this because it seems a file with GeoSPARQL RIF rules, which should probably be part of the profile, is still missing. 
Or perhaps the RIF rules could be described in the profile, without a `prof:hasArtifact` statement?

jabhay:
> Do want to see this PR as a definitive profile for GeoSPARQL 1.0? Or as the start of it, to be open to possible changes later on?
> I am asking this because it seems a file with GeoSPARQL RIF rules, which should probably be part of the profile, is still missing.
> Or perhaps the RIF rules could be described in the profile, without a `prof:hasArtifact` statement?

The latter, the start of it that will change to reflect more outputs that this working group creates that relate to the 1.0 standard. e.g., the RIF rules you refer to, which are part of another issue.

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+
+<http://www.opengis.net/def/geosparql>
+    a prof:Profile ;
+    dcterms:identifier "geosparql" ;
+    dcterms:created "2009-10-27"^^xsd:date ;
+    dcterms:modified "2012-07-19"^^xsd:date ;
+    dcterms:issued "2020-xx-xx"^^xsd:date ;
```
FransKnibbe, line 15: I believe 2020-xx-xx is not as valid xsd:date. If only the year is known, why not use xsd:gYear?

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+    dcterms:publisher <https://www.ogc.org> ;
+    dcterms:title "GeoSPARQL Specification"@en ;
+    dcterms:description "This profile supports representing and querying geospatial data on the Semantic Web. GeoSPARQL defines a vocabulary for representing geospatial data in RDF, and it defines an extension to the SPARQL query language for processing geospatial data. It is presented in a series of artefacts: the vocabulary as an RDF ontology and the specification as a PDF document." ;
```
FransKnibbe, line 18: "This profile supports representing and querying geospatial data on the Semantic Web": I would say GeoSPARQL does that, and the profile describes a version of GeoSPARQL as a set of resources.

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
+PREFIX dcterms: <http://purl.org/dc/terms/>
+PREFIX owl: <http://www.w3.org/2002/07/owl#>
+PREFIX prof: <http://www.w3.org/ns/dx/prof/>
+PREFIX role: <http://www.w3.org/ns/dx/prof/role/>
```
FransKnibbe, line 4: URIs using this prefix, e.g. http://www.w3.org/ns/dx/prof/role/scheme, do not resolve.  This makes it hard to check if the assigned roles make sense.

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+PREFIX sdo: <https://schema.org/>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+
+<http://www.opengis.net/def/geosparql>
+    a prof:Profile ;
+    dcterms:identifier "geosparql" ;
```
FransKnibbe, line 12: A `dcterms:identifier` is supposed to be "An unambiguous reference to the resource within a given context". This is ambiguous, because the profile is not the specification and the version identifier is missing.

The examples in https://w3c.github.io/dx-prof/prof use `rdfs:label` instead of `dcterms:identifier.` So I suggest changing this to `rdfs:label "Profile of GeoSPARQL 1.0"@en`.

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+    dcterms:created "2009-10-27"^^xsd:date ;
+    dcterms:modified "2012-07-19"^^xsd:date ;
```
FransKnibbe, line 14: The `dcterms:created` and `dcterms:modified` dates seem to apply to the GeoSPARQL specification itself, not its profile. That doesn't seem right.

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+    dcterms:issued "2020-xx-xx"^^xsd:date ;
+    dcterms:publisher <https://www.ogc.org> ;
+    dcterms:title "GeoSPARQL Specification"@en ;
```
FransKnibbe, line 17: Again, (if I understand correctly) the resource that is described is a profile of GeoSPARQL, not GeoSPARQL itself. So I would expect a title like "Profile of GeoSPARQL 1.0".

FransKnibbe (changes requested):
Finished the review for now. For completeness I would like to check the assigned roles once they can be resolved.

I did not intend to have different kinds of comments. but it seems some are labeled `pending`.  That can be ignored. Sorry about that, I can't find a way to undo the `pending` label.

1.0/profile.ttl
nicholascar: I was to update this when we actually present the Profile, so some time after this PR has gone in. Likely it will be 2021-xx-xx so I will indeed update it to "2021"^^xsd:gYear to keep it valid! 

I wanted this date separate from the created and modified, just as a testing of DCTERMS' properties.

1.0/profile.ttl
nicholascar: Yes, I have asked the W3C to ensure that this URI resolves as it is actually specified in the Profiles Vocabulary [1] but there is an error with their URI hadling.

[1] https://www.w3.org/TR/dx-prof/#Role:schema

1.0/profile.ttl
nicholascar: Absolutely correct: a copy and paste error. I have updated the description and it now matches that of the 1.1 profile too which was formulated along the lines you suggest. See Commit https://github.com/opengeospatial/ogc-geosparql/pull/58/commits/b05e95fb7d392cee3519ad5389416d4d7f0bfd9a

1.0/profile.ttl
nicholascar: The profiles do also have `dcterms:title` values abd the specific reason for the `dcterms:identifier` property was for use in cataloguing but perhaps a version IRI is needed since we will have 2 profiles for GeoSPARQL, 1.0 & 1.1, with potentially the same IRI. So I've removed `dcterms:identifier` and added `owl:versionIRI`.

1.0/profile.ttl
nicholascar: Yes, right again @FransKnibbe. I've changed these properties

1.0/profile.ttl
```diff
@@ -0,0 +1,63 @@
...
+        <http://www.opengis.net/ont/geosparql> ,
+        <http://www.opengis.net/def/geosparql/funcsrules> ,
+        <http://www.opengis.net/ont/sf> ;
+.
+
+<https://www.ogc.org>
+    a owl:NamedIndividual ;
+    a sdo:Organization ;
+    sdo:name "Open Geospatial Consortium" ;
+    sdo:url <https://www.ogc.org> ;
```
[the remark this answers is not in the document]
nicholascar: It does two things: it establishes that the OGC is an Agent (or the derived `sdo:Organization`) and we just happen to use the same URI they use for their public website as our identifier for them, which is reasonable since we have no other sensible identifier, but also I'm hoping the OGC and others *will* publish stable Linked Data identifiers for themselves eventually. 

So, ideally we would use a better PID for the OGC and then link that to their website via `sdo:url`. Any suggestions for their PID? I have considered using a ROR, in this case <https://ror.org/00nz0j151> (see https://ror.org/search?query=open+geospatial+consortium).

1.0/profile.ttl
nicholascar: Updated as you suggest. Sorry, it does look like I did reuse a lot of properties from GeoSPARQL ont without sufficint modification...

1.0/profile.ttl
nicholascar: Updated to 2021 gYear

1.0/profile.ttl
```diff
@@ -0,0 +1,64 @@
...
+    dcterms:created "2020-11-20"^^xsd:date ;
+    dcterms:modified "2020-12-19"^^xsd:date ;
+    dcterms:issued "2021"^^xsd:gYear ;
+    dcterms:publisher <https://www.ogc.org> ;
+    dcterms:title "GeoSPARQL 1.1 Profile" ;
+    dcterms:description """This is a 'profile declaration' for the GeoSPARQL 1.0 specification (standard) as originally published in 2012. It describes the multiple parts of the specification and how the standard relates to other standards. It is formulated according to the Profiles Vocabulary."""@en ;
+    skos:scopeNote """Profile declarations are used to allow data to identify specifications in a way that allows data to make conformance claims to them - a broader conformance claim than that made to individual 'conformance classes' that are commonly found in recent OGC specifications. Profile delcarations are also used to describe and relate the often multiple parts of specifications: ontologies, specification 'documents', vocabularie, validators and so on. This declaration describes where those parts are, what form (format) they take, what information models they implement and what roles they play."""@en ;
```
FransKnibbe, line 19: Two typos: delcarations and vocabularie

1.0/profile.ttl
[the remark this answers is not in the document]
FransKnibbe: How about http://www.opengis.net/def/ogc? Wouldn't it stand to reason to have the OGC Naming Authority define the OGC?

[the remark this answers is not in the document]
dr-shorthair: https://www.wikidata.org/entity/Q776903 
https://ror.org/00fsdxs93 is a better ROR IRI

1.0/profile.ttl
nicholascar: Fixed in Commit https://github.com/opengeospatial/ogc-geosparql/pull/58/commits/58677653615b875e834f6eb1661933c50dadcf59

1.0/profile.ttl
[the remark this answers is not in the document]
nicholascar: I've added Simon's suggested ROR URI for the OGC in Commit https://github.com/opengeospatial/ogc-geosparql/pull/58/commits/12ef349a1b6bbb84d5b7f1f2d7fd5193053c8728

I think this will have to do for now (internal SWG review). We can put it to the Naming authority for them to correct. They will be the ones who decide on the question _Wouldn't it stand to reason to have the OGC Naming Authority define the OGC?_

1.0/profile.ttl
```diff
@@ -0,0 +1,64 @@
...
+
+
+<http://www.opengis.net/def/geosparql>
+    a prof:Profile ;
+    owl:versionIRI <http://www.opengis.net/def/geosparql/1.0> ;   
+    dcterms:created "2020-11-20"^^xsd:date ;
+    dcterms:modified "2020-12-19"^^xsd:date ;
+    dcterms:issued "2021"^^xsd:gYear ;
+    dcterms:publisher <https://ror.org/00fsdxs93> ;
+    dcterms:title "GeoSPARQL 1.1 Profile" ;
```
FransKnibbe, line 17: should be:  `dcterms:title "GeoSPARQL 1.0 Profile" ;`

nicholascar: Fixed with Commit https://github.com/opengeospatial/ogc-geosparql/pull/58/commits/a571175864dae1477076ca0c197bae2688ced0e1

FransKnibbe (approved):

FransKnibbe:
It all looks good to me now. This should be a good start for continuing to keep track of all versions and related resources.

1.0/profile.ttl
```diff
@@ -0,0 +1,64 @@
...
+PREFIX role: <http://www.w3.org/ns/dx/prof/role/>
+PREFIX sdo: <https://schema.org/>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+
+<http://www.opengis.net/def/geosparql>
+    a prof:Profile ;
+    owl:versionIRI <http://www.opengis.net/def/geosparql/1.0> ;   
+    dcterms:created "2020-11-20"^^xsd:date ;
```
dr-shorthair, line 13: I'm a little unconformable with the fact that the actual date of the GeoSPARQL 1.0 is not in its own property. 
I can see that the profile resource was only created now, but the things that it describes date from 2012.

nicholascar: > "the actual date of the GeoSPARQL 1.0"

You are seeing the date of the _GeoSPARQL 1.0 Profile_, not the date of the thing most commonly referred to as "GeoSPARQL" which would be the specification (HTML doc) and/or the ontology, which have the 2012 date.

One of the things these profile declarations will do is to sort out the ambiguity about what is refereed to by "GeoSPARQL".

dr-shorthair: Yes, I said that. But I raised an important related issue.

nicholascar: So do you propose a change to this PR?

nicholascar: (these profile declarations are expected to change or be edited throughout this SWG so you could always propose a change or even just raise an Issue here and we will deal with it after the merger of this PR)

dr-shorthair: It feels to me like the profile (understood as a suite of artefacts) was implicitly or conceptually created in 2012. The fact that this particular formalization of the description was only created now is of little interest.

nicholascar: That is what I originally indicated with a profile date of 2012 - same as the spec doc/ontology. Feedback from Franz was to have the Profile use a separate date (now).

I don't think we have a strong answer for this point. Perhaps we can lean on `dcterms:issued` and therefore have something like:

- `dcterms:created` - implicit creation date of the profile, ~2012
- `dcterms:issued` - unveiling of profile, now
- `dcterms:modified` - latest edit date of profile

The first line of the profile description - "This is a 'profile declaration' for the GeoSPARQL 1.0 specification (standard) as originally published in 2012." - I think will support this.

Would you agree to this change?

dr-shorthair: `created` vs `issued` might be a pragmatic and usable pattern. I'm not quite as confident about the semantics of `modified` though. But would like to have some other opinions. 

Note that my original comment was _not_ a critique of the use of the dcterms predicates as shown. It _was_ that the date '2012' was not available discretely (yes, I noticed it embedded within a string literal, but that's not really good enough, eh?). This meta-level stuff is always a bit tricky.

nicholascar: OK, I've added in the creation date as suggested above for now.

FransKnibbe: Did the concept of a profile declaration really exist in 2012? We shouldn't practice revisionism.
Furthermore, the definition of `dcterms:created` is "Date of creation of the resource". With  `dcterms:created` in this case being a property of the profile web resource, it seems like we are saying the profile as a web resource was created in 2012.

Would it help to assign dates to the individual resources making up the profile declaration? E.g. indicate that the GeoSPARQL documentation and ontology where published in 2012? That should clarify matters regardless of the dates assigned to the profile. 

In the line  "This is a 'profile declaration' for the GeoSPARQL 1.0 specification (standard) as originally published in 2012." I read 2012 as applying to the specification, not to the profile declaration.

nicholascar: On review, I think I'm not in favour of using created date of 2012 here since, as @FransKnibbe says, there was no notion of a profile back then. So I'm reverting my creation date of the _profile_ to now and the resources within the profile will have to stand on their own two feed with their 2012 creation dates.

nicholascar: ...and with this I'm merging since I have two approvals here and this is a small issue that, if necessary, can be picked up again in subsequent PRs.

Issue #44, opened 2020-10-04
GeoSPARQL and Common Query language CQL


situx:
The common query language (CQL) (http://docs.opengeospatial.org/DRAFTS/19-079.html#simple-cql_filter-expression) is used in many geospatial web services (WFS) and in the currently in OGC API features to filter geospatial featurecollections.

In one of our research projects we developed a "semanticwfs" (https://github.com/i3mainz/semanticwfs), i.e. a WFS/OGC API Features compatible webservice with triple store backends which has to convert CQL filter expressions to (Geo)SPARQL internally.
Test Instance here: https://www.i3mainz.de/projekte/bkg/semanticwfs/

For this usecase (using a GeoSPARQL compliant triple store as a backend of an OGC API Features compatible web service) mappings from CQL to GeoSPARQL have to be defined.

If this is a usecase which is interesting to follow up we could define the new GeoSPARQL version keeping CQL compatibility in mind, as OGC API Features and WFS are still the main form of accessing geospatial data for the GIS community.

I am just putting this out there for discussion as I am not sure if it is something we should consider or if this is the right place.

nicholascar:
Hi @situx, I've just implemented a series of what I call _OGC LD APIs_, which are OGC API instances implemented using a Linked Data API framework - [pyLDAPI](https://github.com/rdflib/pyLDAPI). The back end for 2 of them is a triplestore with GeoSPARQL data in it:

1. http://asgs.surroundaustralia.com/
2. http://geofabric.surroundaustralia.com/

Another one has a WFS back-end:

3. http://provinces.surroundaustralia.com/

I guess I've taken a related approach to your semanticwfs but my implementations are both OGC APIs and Linked Data APIs in the same place!

These are all test systems and one of the things they need to move to more solid implementations is better filtering- currently they only have BBOX filters in place.

So I'm interested in CQL-like filters or CQL-to-(Geo)SPARQL.

Happy to keep chatting about this whether or not it actually goes into GeoSPARQL 1.1.

_(nice cuneiform GitHub image by the way)_

situx:
Hi @nicholascar,

great to see there are similar approaches in this direction.
We did not consider a Linked Data API implementation for our project, but it seems like the next logical step and should also not be hard to implement. I will discuss it.

[...]

PR #68, opened 2020-12-23
Relation of CQL to GeoSPARQL

situx:
I have started to work on issue #44  and hope that this is a good place to start.
I am happy to receive further input, corrections and contributions.


nicholascar:
@situx is this ready for review or do you want to talk through it at the next meeting (in about 13 hours from now)?

situx:
No this is not ready at all. It was a start and needs more contributions

situx:
If it is just about the tables, we could consider merging them into GeoSPARQL 1.1, if we want a more comprehensive documentation on how to use a GeoSPARQL compatible triple store as a backend of an OGC API Features service then this pull request is not ready. 
Any opinions?

1.1/spec/18-Annex-C.adoc
```diff
@@ -0,0 +1,129 @@
+=== Annex C 
+=== (informative)
+=== Best-practice: A GeoSPARQL-enabled triple stores as a backend of an OGC web service
+
+This annex wants to present best practices of creating linked data backends to OGC webservices.
+
+==== C.1 Accessing geospatial features in a SPARQL endpoint
+
+Geospatial features in SPARQL endpoints as defined in the GeoSPARQL standard are instances of subclasses of http://www.opengis.net/ont/geosparql#Feature[geo:Feature].
+They include one or many http://www.opengis.net/ont/geosparql#hasGeometry[geo:hasGeometry] relations and a set of further properties which may or may not refer to further instances related to the geospatial feature.
```
dr-shorthair, line 10: > and a set of further properties which may or may not refer to further instances related to the geospatial feature. 

Maybe just: 

> and further properties related to the geospatial feature.

1.1/spec/18-Annex-C.adoc
```diff
@@ -0,0 +1,129 @@
...
+  ?item rdf:type ?class .
+  ?class rdfs:subClassOf* geo:Feature .
+  ?item ?rel ?val .
+  OPTIONAL {?item geo:hasGeometry/geo:hasSerialization ?geo .}
+}
+```
+This SPARQL query retrieves all properties and their values directly connected to an instance of all classes which are subclasses of http://www.opengis.net/ont/geosparql#Feature[geo:Feature].
+The query result is a list of triples which may be converted to a FeatureCollection representation.
+
+This representation results in a valid FeatureCollection which may be exposed using an OGC API Features compatibles webservice.
```
dr-shorthair, line 25: > compatibles

Maybe delete? (no s anyway)

1.1/spec/18-Annex-C.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+However, it needs to be remarked that the aforementioned query however may not in all cases deliver the the expected result.
```
dr-shorthair, line 27: > it needs to be remarked that the aforementioned

--> 

'this'

dr-shorthair (changes requested):

situx:
Ok @dr-shorthair I changed the things you requested

dr-shorthair (approved):

nicholascar:
I've just made a bunch of small updates, mainly to introduce the `geo:FeatureCollection` class into this Annex. This might need more work to fully run the example through, however likely we can't complete this until the `geo:FeatureCollection` PR (#150) is completed.

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
 === Annex C 
 === (informative)
 
+=== Best-practice: A GeoSPARQL-enabled triple store as a backend of an OGC web service
+
+This annex presents best practices for the creation of Linked Data backends to OGC webservices.
+
+==== C.1 Accessing geospatial features in a SPARQL endpoint
```
FransKnibbe, line 8: Why "geospatial", not "spatial"? GeoSPARQL features are not inherently geographical.

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+
+Geospatial _features_ accessed via SPARQL endpoints <<SPARQLPROT>> are, as defined in the GeoSPARQL standard, instances of http://www.opengis.net/ont/geosparql#Feature[geo:Feature] or of subclasses of it.
+They may include one or more http://www.opengis.net/ont/geosparql#hasGeometry[geo:hasGeometry] relations and further properties related to the feature and may be grouped into http://www.opengis.net/ont/geosparql#FeatureCollection[geo:FeatureCollection] instances where `geo:FeatureCollection` is a new class in GeoSPARQL 1.1 specifically for the description of collections of `geo:Feature` instances.
```
FransKnibbe, line 11: feature → Feature

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
 === Annex C 
 === (informative)
 
+=== Best-practice: A GeoSPARQL-enabled triple store as a backend of an OGC web service
+
+This annex presents best practices for the creation of Linked Data backends to OGC webservices.
+
+==== C.1 Accessing geospatial features in a SPARQL endpoint
+
+Geospatial _features_ accessed via SPARQL endpoints <<SPARQLPROT>> are, as defined in the GeoSPARQL standard, instances of http://www.opengis.net/ont/geosparql#Feature[geo:Feature] or of subclasses of it.
```
FransKnibbe, line 10: GeoSPARQL Features are not necessarily geographical.

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+They may include one or more http://www.opengis.net/ont/geosparql#hasGeometry[geo:hasGeometry] relations and further properties related to the feature and may be grouped into http://www.opengis.net/ont/geosparql#FeatureCollection[geo:FeatureCollection] instances where `geo:FeatureCollection` is a new class in GeoSPARQL 1.1 specifically for the description of collections of `geo:Feature` instances.
+
+The following example SPARQL query retrieves all feature collections of a given SPARQL endpoint. A third-party webservice may convert the query results to one or many geoformat serializations.
```
FransKnibbe, line 13: feature collections → Feature Collections (class names are capitalised).

FransKnibbe: I would prefer just "data formats" or something like that instead of "geoformat serializations"

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+```turtle
+SELECT ?item ?rel ?val ?geom 
+WHERE {
+  ?item rdfs:subClassOf* geo:Feature .
+  ?item ?rel ?val .
+  OPTIONAL {?item geo:hasGeometry/geo:hasSerialization ?geom}
+}
+```
+
+This SPARQL query retrieves all features' properties and their values directly connected to an instance of all classes which are subclasses of http://www.opengis.net/ont/geosparql#Feature[geo:Feature].
```
FransKnibbe, line 39: feature → Feature

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+The query result is a list of triples which may be converted to a `geo:FeatureCollection` instance.
+
+`geo:FeatureCollection` match the OGC API Features's <<OGCAPIF>> concept of a FeatureCollection and thus may be exposed using an OGC API Features webservice.
```
FransKnibbe, line 42: match → matches

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+
+However, it needs to be remarked that this may not deliver the expected result in all cases. There are several reasons for this:
```
FransKnibbe, line 44: "it needs to be remarked that this may not deliver" can be shortened to "this may not deliver"

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+
+- More information for the `geo:FeatureCollection` instance may be needed, in particular, an identifier and perhaps a label for it. If the back-end data store also contains information for the `geo:FeatureCollection` instance then this may be queried for to solve this issue. If not, the API might need to create such information
```
FransKnibbe, line 46: information → data

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+- Maybe only information in a certain namespace is of interest. The solution is to apply FILTER expressions to the SPARQL query
```
FransKnibbe, line 47: information → data

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+- The above query only queries directly connected properties and their values. Often geodata formats like GML are hierarchical and therefore need further resolution of property values depicted as URIs. If that is the case the query needs to be adjusted.
+
+==== C.2 Mappings from CQL statements to GeoSPARQL queries
+
+This section presents lists of equivalences between CQL statements and GeoSPARQL statements.
```
FransKnibbe, line 52: This seems like a good place to explain what CQL means and to refer to its specification document

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+| EQUALS(geometry1,geometry2)  | FILTER(http://www.opengis.net/def/function/geosparql/sfEquals[geof:sfEquals](?geometry1,?geometry2))  
+| INTERSECTS(geometry1,geometry2)  | FILTER(http://www.opengis.net/def/function/geosparql/sfIntersects[geof:sfIntersects](?geometry1,?geometry2)) 
+| OVERLAPS(geometry1,geometry2)  | FILTER(http://www.opengis.net/def/function/geosparql/sfOverlaps[geof:sfOverlaps](?geometry1,?geometry2))  
+| TOUCHES(geometry1,geometry2)  | FILTER(http://www.opengis.net/def/function/geosparql/sfTouches[geof:sfTouches](?geometry1,?geometry2))  
+| WITHIN(geometry1,geometry2)  | FILTER(http://www.opengis.net/def/function/geosparql/sfWithin[geof:sfWithin](?geometry1,?geometry2))  
+|======================
+
+===== C2.6 Temporal Operators
+
+Temporal operators are currently not part of the GeoSPARQL standard.
```
FransKnibbe, line 128: I would remove "currently", to not give the expectation of GeoSPARQL becoming temporal one day.

FransKnibbe (commented):
I have three further general comments:
1) Formatting of class names is not consistent. It would be good to use the same guidelines used in [PR 186](https://github.com/opengeospatial/ogc-geosparql/pull/186).
2) Are we sure this annex describes best practices? For something to be a best practice there have to be many or several practices (i.e. not thought experiments) and a method of separating the bad, good and best practices.
3) Is an annex of the specification really the best place to describe a best practice? I would sooner expect such a thing to be a separate document. Or maybe a section in https://www.w3.org/TR/sdw-bp/. If it were to be a separate document, it could also be made easier to update once in a while (not requiring the same processes as an OGC standard)

situx:
Hi @FransKnibbe thanks for your comments. I will go through them and fix them now. 
As for the questions about whether best practice is a good term we could instead use recommendations.
As for the question about if it is correct to put it in this annex:
Is does not need to be there if people think that there is a better place for that e.g. in the spatial data on the web best practices document.
I am open to that. Whereever it fits best.

situx:
Ok, I addressed the comments of @FransKnibbe except for the decision about best practices and where to publish this document

1.1/spec/18-Annex-C.adoc
```diff
@@ -68,3 +211,8 @@ The following table maps the functions and properties from Simple Features for S
 | Centroid(): Point | http://www.opengis.net/def/function/geosparql/centroid[geof:centroid] | 1.1 | http://www.opengis.net/ont/geosparql#hasCentroid[geo:hasCentroid] | 1.1
 | PointOnSurface(): Point | N/A | - | N/A | -
 |===
+
+
+==== References
```
nicholascar, line 216: Can you perhaps just place this Reference in the main [Bibliography](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/spec/19-Bibliography.adoc)? We don't have separate References sections for other sections.

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+
+- Filter CRS parameter equivalent does not yet exist in GeoSPARQL?
+
+===== C2.5 Geospatial Operators
+
+GeoSPARQL includes equivalents of many CQL filter functions as can be seen in the table below.
+
+|======================
+| CQL Filter Expression  | GeoSPARQL Filter Function 
+| CONTAINS(geometry1,POINT(1 1))  | FILTER(http://www.opengis.net/def/function/geosparql/sfContains[geof:sfContains](?geometry1,"POINT(1 1)"^^geo:wktLiteral))
```
nicholascar, line 116: Why have you used a literal instead of a variable for CONTAINS()? Can you just say `CONTAINS(geometry1,geometry2)` just like the others?

I don't think GeoSPARQL `sfContains()` is limited to a `POINT` for the second parameter.

nicholascar:
> As for the question about if it is correct to put it in this annex:

I'm happy for this to be in an Annex and I will be able to supply a full implementation of it in our Linked Data API within a month or so. We have productions APIs like http://asgs.surroundaustralia.com/ which have a partial CQL implementation under the hood but which has been hidden for now. With this Annex appearing in the Spec, we'll complete the implementation.

1.1/spec/18-Annex-C.adoc
situx: No it is not. It just followed the CQL Filter expression I used as an example. I could change it to use variables only, but the CQL Filter expressions do not allow the representation of variables as a second parameter. I wanted to express an equivalent representation of exactly the CQL statement.

1.1/spec/18-Annex-C.adoc
situx: Yes, I will change it accordingly

FransKnibbe:
> I'm happy for this to be in an Annex

But is it the right place? To me it does not seem like a core element of the standard. It is a useful guide to implementation of GeoSPARQL in certain scenarios. So it would seem more appropriate to publish it as a separate but linked document. I notice that the OGC maintains a collection of best practice documents: [https://www.ogc.org/docs/bp](https://www.ogc.org/docs/bp). Wouldn´t the document be more at place there?
One consideration is that more practical guides for putting GeoSPARQL to use can be made in the future. Should they then be added as annexes to the standard too?
Also, is it really the most _convenient_ to have this best practice guide be part of the core standard? That will make it harder to update the document when necessary.

pvretano:
Just a heads up ... the current version of [CQL](https://docs.ogc.org/DRAFTS/19-079r1.html) was rebranded as CQL2.  The reason for this ... CQL was originally derived for the OGC Catalogue specification.  CQL was badly specified in OGC CAT so we cleaned up the original BNF to fully specify the language.  For a long time this reworked CQL was backward compatible with the CQL from OGC CAT.  Over time, however, based on feedback from implementers the two drifted apart enough that we felt that we should rebrand CQL to CQL2.

situx:
Ok, I moved the references, added a new reference to CQL2 and renamed all occurrences of CQL to CQL2

1.1/spec/18-Annex-C.adoc
nicholascar: > the CQL Filter expressions do not allow the representation of variables as a second parameter

I didn't mean that: I wanted to make sure that the second parameter could be a POLYGON instead of a POINT. Hard coded as you have it makes it look like a user can only test for a POINT inside something (1st param) but it should be test for anything inside anything else.

situx: Ok, changed it to use the variable name instead

1.1/spec/18-Annex-C.adoc
```diff
@@ -1,6 +1,149 @@
...
+| number BETWEEN 5 AND 10 | ?item my:number ?number . FILTER(?number>=5 && ?number<=10) | BETWEEN statements are converted to arithmetic expressions
+| name IN ("OGC","W3C") | VALUES ?namevalues {"OGC" "W3C"} ?item my:name ?namevalues . | IN statements are expressed by VALUES statements
+| name IS NOT NULL | EXISTS {?item my:name ?name } | NOT NULL statements are converted to EXIST statements
+| name LIKE "OGC." | ?item my:name ?name . FILTER(regex(?name, "OGC.", "i" )) | LIKE statements are converted to SPARQL regex filters
+| INTERSECTS(geometry1, geometry2) | FILTER(http://www.opengis.net/def/function/geosparql/sfIntersects[geof:sfIntersects](?geometry1,?geometry2)) | The INTERSECTS filter statement is converted to a GeoSPARQL FILTER statement
+|======================
+
+- Filter CRS parameter equivalent does not yet exist in GeoSPARQL?
+
+===== C2.5 Geospatial Operators
```
nicholascar, line 110: Rename to just _Spatial Operators_ since that's the terminology CQL2 uses. Also, these are generic spatial, not just geospatial, operators!

situx: Done

situx:
Ok, as we decided in the last call that this document shall be separate from the standard, I reintegrated the references and put it in a standalone folder.
I will merge this version. This document may further be extended without the need to be coupled to a GeoSPARQL release as I understand it.

PR #70, opened 2021-01-04
1.0 rules in RIF + profile update

### LINKED ISSUE #67: Find RIF rules for the 1.0 Profile and present it as a profile Resource
The RIF rules might be described and even templated within the ontology but they are not available via a stand-alone asset. They should be extracted and placed into an asset (RIF document) described within the Profile.

nicholascar:
RIF rules implements for GeoSPARQL 1.0 in RIF Presentation Format using the template in GeoSPARQL 1.0, Clause 11.

Closes #67


1.0/rules.rifps
```diff
@@ -0,0 +1,1686 @@
...
+    Group (
+        # geo:sfEquals
+        Forall ?f1 ?f2 ?g1 ?g2 ?g1Serial ?g2Serial (
+            ?f1[geo:sfEquals->?f2] :-
+            Or (
+                # feature – feature rule
+                And (
+                    ?f1[geo:hasDefaultGeometry->?g1]
+                    ?f2[geo:hasDefaultGeometry->?g2]
+                    ?g1[geo:gmlLiteral->?g1Serial]
```
mperry455, line 14: I think we need the property geo:asGML here instead of the datatype geo:gmlLiteral. So it would be ?g1[geo:asGML->?g1Serial]. The same thing goes for geo:asWKT and geo:wktLiteral.

We have a bit of a combinatorial explosion here because the functions like geof:sfEquals should work for mixed input - that is, comparing a WKT geometry with a GML geometry. We would need all these rules for (WKT, WKT) and (WKT, GML), (GML, WKT), (GML, GML). This becomes an even bigger problem when GeoJSON and KML are added. As an alternative, we could use the generic geo:hasSerialization property instead. geo:asWKT and geo:asGML are subPropertyOf geo:hasSerialization. This requires some RDFS reasoning support in the implementation though.

situx: It should be possible to create all permutations automatically. We could write a small script to do this and if more literal types are added just call the script with the newly available literal types.

situx: I wrote a script in #76 that would solve this issue, in case we would like to use RIF rules only and not rely on reasoning.

nicholascar: Of course I produced the RIF rules doc in this PR  via a script too! I've added my scripts details to PR #76 and we will likely be able to produce a combined script for the profiles (it should be made available in both the 1.0 and 1.1 profiles with different inputs).

Once the script is finalised, I will ensure it is listed in the profiles correctly.

situx: Great! I will try to merge the script!

nicholascar:
I have updated the RIF rules using the updated script in PR #76 and included the script here, as well as its ResourceDescriptor in the profile

nicholascar:
@mperry455 @situx if you've approved PR #76, you should approve this one too!

situx (approved):

mperry455 (approved):

Issue #69, opened 2021-01-04
Support KML literals in GeoSPARQL


nicholascar:
As per the note in GeoSPARQL 1.0 Sec. 12:

> ...Obvious extensions are to define new conformance classes for other standard serializations of geometry data (e.g. KML, GeoJSON). ...

We should add support for KML literals for completeness. Given the added support for GeoJSON, this should be an extremely simple task.

A small review of the status of KML should be conducted and used to justify/exclude this task.

situx:
I have implemented KML literals before. I think it should be straightforward to do. 
It would need the same restriction as GeoJSON that only WGS84 is a valid CRS.
Apart from that, I see no reasons to not include it. In fact, I think as an OGC standard I would find it logical to support at least all standards defined by the OGC as literals.

nicholascar:
> I would find it logical to support at least all standards defined by the OGC as literals.

Yes, we should! Happy for you to put this through- I have a lot of other GeoSPARQL things on...

situx:
Since I had some time just now, I made a draft in #75 
We can discuss whether or not to include it in the next meeting.

situx:
Closing this issue as the corresponding pull request #75  has been accepted and closed

PR #75, opened 2021-01-04
KML Literal definition

situx:
This definition if approved closes issue #69 

KML like GeoJSON requires a WGS84 coordinate reference system for every geometry to be defined.

KML, as opposed to other literals, does not define an empty geometry.
It requires at least one coordinate to be present in the coordinate tag of every geometry according to the KML specification. 

However, we may define an empty KML geometry as the empty literal (as proposed in my draft of the specification). 

Apart from that, I see no issues in including KML literals into GeoSPARQL.


dr-shorthair (approved):

nicholascar (approved):
@situx with two Approval reviews you should now merge this in!

PR #97, opened 2021-02-27
Wkt uri spec

nicholascar:
Updates the WKT literal spec with ABNF format indication, as per the DGGS WKT literal.

Also adds a number of missing references and changes all mention of URI to IRI to align all references to IRI/URI.

All edits are in Part 8 other than the URI -> IRI replacements which are in multiple files.


1.1/spec/06-Part-03.adoc
```diff
@@ -6,7 +6,7 @@ ISO 19125-1, _Geographic information — Simple feature access — Part 1: Commo
 
 OGC 07-036, _Geography Markup Language (GML) Encoding Standard_, Version 3.2.1
 
-RFC 2396, _RFC 2396 - Uniform Resource Identifiers (URI): Generic Syntax_
+RFC 2396, _RFC 2396 - Uniform Resource Identifiers (IRI): Generic Syntax_
```
akuckartz, line 9: Something is wrong here. Likely a different RFC should be referred to.

akuckartz:
I am not convinced that replacing URI by IRI is really helpful or necessary.

1.1/spec/06-Part-03.adoc
nicholascar: I've fixed this in the commit below by replacing the reference to the URI spec with a reference to the IRI spec.

I will update the _Normative References_ section of GeoSPARQL in a subsequent PR.

nicholascar:
> I am not convinced that replacing URI by IRI is really helpful or necessary.

I don't think it's _super duper_ useful but I think it's a sensible part of updating all the references in this spec. I do think a few more edits will be necessary to align all references to IRIs, not URIs, as per the fix above relating to your other comment.

Thanks for finding these small annoyances! After the current set of small PRs are merged in I will start an end-to-end review of the whole spec document to fix small errors, align styles and generally improve the ASCIIDoc source and spec presentation.

1.1/spec/06-Part-03.adoc
dr-shorthair: The RFC for IRI is https://tools.ietf.org/html/rfc3987

nicholascar: That's the reference I've used in _Bibliography_ and have just now overwritten the _Normative References_ URI reference with.

(I suspect some of the confusion with this PR and some others is that there's a spillage of concerns between them as I find an issue, address it but then have to address underlying issues - this PR was really only meant to be about WKT! Well, to fix this I am doing that whole-spec review this week.)

situx (approved):

akuckartz:
It might be better to refer to https://url.spec.whatwg.org/ instead which states:
"Standardize on the term URL. URI and IRI are just confusing."

dr-shorthair:
It depends on how it is being used. 'Identifier' is different from 'Locator'. The former is implicitly persistent, while the latter is just the current directions.

nicholascar:
> Standardize on the term URL. URI and IRI are just confusing.

I really don't think we can do this. We really do mean something other than just a URL, as Simon indicates, and I do think we - modern standards makers - should be pushing the envelope a little bit in terms of the underlying standards that are used so I prefer IRI over URI.

I am now reviewing the Specification document as a whole and am adding in a clarification paragraph about URI/IRI in the Introduction where IRI is first mentioned. I will also check with the OGC Naming Authority to see what they recommend.

1.1/spec/11-Part-08.adoc
```diff
@@ -655,14 +660,14 @@ SPARQL language extensions may provide additional associations between operators
...
-Several non-topological query functions use a unit of measure URI. The OGC has defined some standard units of measure URIs under the `+http://www.opengis.net/def/uom/OGC/1.0/+` namespace, for example `+<http://www.opengis.net/def/uom/OGC/1.0/metre>+`.
+Several non-topological query functions use a unit of measure IRI. The OGC has defined some standard units of measure IRIs under the `+http://www.opengis.net/def/uom/OGC/1.0/+` namespace, for example `+<http://www.opengis.net/def/uom/OGC/1.0/metre>+`.
 
 ===== Function: geof:distance
 
 
 ```
 geof:distance (geom1: ogc:geomLiteral, geom2: ogc:geomLiteral, 
-               units: xsd:anyURI): xsd:double
+               units: xsd:anyIRI): xsd:double
```
dr-shorthair, line 670: global search-and-replace has bitten here. It is `xsd:anyURI`.

dr-shorthair (changes requested):
Revert to `xsd:anyURI`

1.1/spec/11-Part-08.adoc
nicholascar: Done!

dr-shorthair (approved):
All good now.

PR #106, opened 2021-03-10
twicethrough

### LINKED ISSUE #90: Fix figure annotations in 8.18.1. B.1 Example Data
```
[RDF Triple] | img/03.png
Figure 3. Illustration of spatial data
```

`RDF Triple]` etc should not be visible

### LINKED ISSUE #95: Update the CHANGELOG in the spec
The current CHANGELOG is as per the original 1.0 spec and needs updating for 1.1.

Consider removing all the 1.0 changes or creating a separate table for 1.1 changes.

### LINKED ISSUE #114: Add a list of 1.1 New Features to the Revision history section
_Revision history_ lists small changes but a major changenote from v 1.0 - 1.1 should be included for quickly understanding the differences and as a checklist against the SWG's Charter and issues for 1.1.

nicholascar:
A second full pass through the Spec to fix many small things.

The resulting Spec:

* [Rendered in HTML](https://raw.githack.com/opengeospatial/ogc-geosparql/twicethrough/1.1/spec/twicethrough.html)

Fixes #114 - commit https://github.com/opengeospatial/ogc-geosparql/pull/106/commits/720e95e9130534170dfb5ee0bbb5364a66b941ca

Fixes #95 - commit https://github.com/opengeospatial/ogc-geosparql/pull/106/commits/720e95e9130534170dfb5ee0bbb5364a66b941ca

Fixes #90 - https://github.com/opengeospatial/ogc-geosparql/pull/106/commits/39ada04ca5466cc5ab044bfba32a81874031433a


situx:
I have a question about the numbering of requirements:

Does it make sense to number requirements in an ascending manner as it was done in the last spec, or should we create numbers according to categories?

For example: 

We have three requirements for WKT serializations, in fact, we will for every literal in the future have at least three requirements.
Would it make sense to number these as e.g. req 15.1, 15.2, 15.3 or shall we stick to 15,16,17?

nicholascar:
Yes I think section or some other such grouping of Requirements it needed. We need to see how the Requirements align with the conformance classes. If we can tabulate Conformance Classes & Requirements, then we can probably use a 2-part numbering like Requirement {CC}.{R} (CC: Conformance Class number, R: Requirements number).

I think that re-numbering should be a separate PR from this one though as it will need a couple of reviews from @mperry455 and also perhaps other OGC Spec editors, such as the OGC API: Features etc. so we can ensure we are doing things in a similar way so as not to confuse people.

FransKnibbe:
I understand the need for grouping, but I wonder if a numbering scheme is the best solution.

With further extensions of the standard coming up, a numbering scheme that looks nice now may become skewed in the future. Besides, sequential numbering is simple, and simple is (often) good.
 
Perhaps we can think of other ways of grouping requirements that are both future-proof and backward compatible?
One way could be to introduce text labels for each requirement, indicating group membership (e.g. "WKT - axis order", "WKT - empty"). Another way could be to publish requirements as a SKOS concept scheme, with arbitrary grouping. This will give each requirement an IRI, which can be referenced in the documentation. And the concept scheme (with requirements as Web resources) could be useful by itself, for instance in conformance testing.

Whichever way requirements are grouped, we need to think of sensible grouping characteristics. I think at least module membership should play a role, because modules are the main units that implementations do or don't support. (see also [issue 120](120)).

nicholascar:
I was going to keep going with this PR but since there are others piling in (thanks @FransKnibbe & @situx) this better be merged sooner rather than later in order to avoid messy merges.

I'll start a fresh review if/when this PR and the examples one goes through.

situx (approved):
I am ok with the changes, however, the Standard Properties for SpatialMeasure and SpatialObject are still missing. They will probably be added in a next pull request?

nicholascar:
> I am ok with the changes, however, the Standard Properties for SpatialMeasure and SpatialObject are still missing. They will probably be added in a next pull request?

Yes, to be done. These changes should be merged (or changes requested) as long as these changes are ok, in and of themselves, but that doesn't mean there aren't loads of other changes still needed!

> Another way could be to publish requirements as a SKOS concept scheme, with arbitrary grouping.

While the semantic engineer in me likes this, we have to get the Spec document presented as best we can too!

> Whichever way requirements are grouped, we need to think of sensible grouping characteristics. 

I agree, we do need to work on this still! Let's have a PR that's just about Requirements after this one, not blended in with other things like this. I confused myself with all the issues dealt with!

1.1/geo.ttl
```diff
@@ -12,41 +12,63 @@
...
-		sdo:email "simon.cox@csiro.au" ;
-		sdo:url <https://orcid.org/0000-0002-3884-3420> ;
-	] ;
+		sdo:email "simon.cox@csiro.au"^^xsd:anyURI ;
+		sdo:url "https://orcid.org/0000-0002-3884-3420"^^xsd:anyURI ;
+	] ,
+	[
+		a sdo:Person ;
+		sdo:name "Frans Knibbe" ;		
+		sdo:affiliation "CSIRO Australia" ;
```
FransKnibbe, line 68: This line should be removed.

1.1/spec/01-Preamble.adoc
```diff
@@ -73,6 +73,63 @@ to
 | 1.1 Draft | GeoSPARQL 1.1 SWG | All | Addition of GeoSPARQL 1.1 elements
 |===
 
+=== Major changes between versions 1.0 and 1.1
+Version 1.1 of GeoSPARQL was released approximately 9 years after version 1.0. It contains no breaking changes to 1.0, but does contain additions: whole new profile resources, new ontology elements and new functions. The major changes are given in the tables below.
+
+These new profile resources are resources - documents - that are separate to this specification. The new _profile defintion_ lists all the GeoSPARQL 1.1 resources.
```
FransKnibbe, line 79: typo: defintion

1.1/geo.ttl
kjano: @FransKnibbe I think your affiliation should be Geodan, correct?

1.1/spec/01-Preamble.adoc
```diff
@@ -73,6 +73,63 @@ to
...
+
+|===
+|New resource | Location
+
+|Profile definition | http://www.opengis.net/def/geosparql
+|GeoSPARQL Rules in RIF | http://www.opengis.net/def/geosparql-rifrules
+|RDF validation file | http://www.opengis.net/def/geosparql-shapes
+|===
+
+These new new ontology elements and new functions are normatively defined in this specification document.
```
FransKnibbe, line 89: typo: new new

1.1/spec/09-Part-06.adoc
```diff
@@ -80,3 +78,170 @@ geo:Feature a rdfs:Class, owl:Class ;
...
+    skos:scopeNote "The target is a geometry that defines a rectilinear region whose edges are aligned with the axes of the coordinate reference system, which exactly contains the geometry or feature e.g. sf:Envelope"@en ;
+    rdfs:domain geo:Feature ;      
+    rdfs:range geo:Geometry .
+```
+
+GeoSPARQL does not restrict the cardinality of the http://www.opengis.net/ont/geosparql#hasBoundingBox[`geo:hasBoundingBox`] property. A feature may be associated with more than one bounding-box, for example in different coordinate reference systems.
+
+==== Property: geo:hasCentroid
+
+The property http://www.opengis.net/ont/geosparql#hasCentroid[`geo:hasCentroid`] is used to link a feature with a point geometry corresponding with the centroid of its geometry. The centroid is typically used to show location on a low-resolution map, and for some indexing and discovery functions.
```
FransKnibbe, line 156: I suggest changing 'low-resolution map' to 'low resolution image' to avoid GIS bias.

1.1/spec/09-Part-06.adoc
```diff
@@ -80,3 +78,170 @@ geo:Feature a rdfs:Class, owl:Class ;
...
+
+==== Property: geo:hasCentroid
+
+The property http://www.opengis.net/ont/geosparql#hasCentroid[`geo:hasCentroid`] is used to link a feature with a point geometry corresponding with the centroid of its geometry. The centroid is typically used to show location on a low-resolution map, and for some indexing and discovery functions. 
+```turtle
+geo:hasCentroid a rdf:Property, owl:ObjectProperty ;
+    rdfs:subPropertyOf geo:hasGeometry;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "has centroid"@en ;
+    skos:definition "The arithmetic mean position of all the geoetry points of a given feature."@en ;
```
FransKnibbe, line 162: typo: geoetry

FransKnibbe (changes requested):
Love the extra explanations! They are valuable additions.
I spotted a few typos, but it is a good idea to merge this soon, to have an updated baseline (and have smaller change packages afterwards).

1.1/geo.ttl
FransKnibbe: Not any longer, @kjano. I went into self-isolation before it became fashionable. For the moment I am unaffiliated; pursuing a few private interests.

nicholascar: I've updated this

1.1/spec/01-Preamble.adoc
nicholascar: I have removed the double new

1.1/spec/09-Part-06.adoc
nicholascar: Fixed

1.1/spec/09-Part-06.adoc
nicholascar: Updated

nicholascar:
I've addressed all of @FransKnibbe points

FransKnibbe (approved):
OK, let's merge!

PR #108, opened 2021-03-11
Transformation functions

situx:
Referring to yesterday's discussion on transformation functions for literal formats I have created a proposal.
I have defined the following functions:
- geof:asWKT(geom: ogc:geomLiteral): geo:wktLiteral
- geof:asGML(geom: ogc:geomLiteral,versionString: xsd:string): geo:gmlLiteral
- geof:asGeoJSON(geom: ogc:geomLiteral): geo:geoJSONLiteral
- geof:asKML(geom: ogc:geomLiteral): geo:kmlLiteral
- geof:asDggsWkt(geom: ogc:geomLiteral, dggsIri: xsd:anyURI): geo:dggsWktLiteral
Should we follow this approach for any upcoming literal?

Also I defined a function:
- geof:transform(geom: ogc:geomLiteral, srsURI: xsd:anyURI): ogc:geomLiteral
I believe this function is sufficient to enable CRS transformations.


nicholascar:
> I am not sure about the signature of an asDGGSWkt function, so I did not add a function definition for that. Maybe you could look into it?

The conversion from a traditional geometry to a DGGS geometry, or between DGGS geometries, will be DGGS-dependent, i.e. conversion for each DGGS will likely be very different.

I think that in time we are likely to see only a few families of DGGS (3 - 6) and the conversions to/from them and between them become more known, so it's reasonable to define this function and to expect that it will eventually be implemented.

After this PR, I will patch this with another PR.

situx:
Ok, but which parameters should the function have? One for the literal which is converted to dggs and another string for the dggs type?

nicholascar:
```
# to return a geometry as a DGGS WKT value, we only need to pass the function the geometry 
# to convert from and the IRI of the DGGS to convert to
# a geo:dggsWktLiteral is returned or, if the given DGGS IRI is unknown a _not implemented_ 
# or equivalent error must be thrown
geof:asDggsWkt(geom: ogc:geomLiteral, dggsIri: xsd:anyURI): geo:dggsWktLiteral
```

situx:
Ok, cool. I added the function signature and also updated the abstract test suite (hopefully correctly).
So this can be a basis for discussion in the next meeting.

dr-shorthair (approved):

1.1/spec/11-Part-08.adoc
```diff
@@ -797,6 +862,14 @@ geof:minZ (geom: ogc:geomLiteral): xsd:double
 
 Returns the minimum Z coordinate for `geom`.
 
+```
+geof:transform (geom: ogc:geomLiteral, srsURI: xsd:anyURI): ogc:geomLiteral
```
nicholascar, line 866: `srsURI` needs to be `srsIRI`. I would fix this myself but @situx must not have checked the box to allow commits directly to this branch as I don't seem to have privileges to do it.

1.1/spec/11-Part-08.adoc
```diff
@@ -797,6 +862,14 @@ geof:minZ (geom: ogc:geomLiteral): xsd:double
...
+```
+
+http://www.opengis.net/def/function/geosparql/transform[geof:transform] converts `geom` to a spatial reference system defined by srsURI. The function raises an error if a transformation is not mathematically possible.
```
nicholascar, line 869: `srsURI` --> `srsIRI`

nicholascar (changes requested):
Thanks @situx. Just two minor changes both for `srsURI` --> `srsIRI`.

situx:
@nicholascar I changed srsURI to SRSIRI. Not sure how I can activate that others can commit to the branch.

nicholascar (approved):
When you create a PR, you can tick some box that says something like "allow others to commit directly to this PR". I'll find the exact wording for you when I next create a PR!

Issue #105, opened 2021-03-10
Update geoaf reference needed


nicholascar:
The Requirement 33 for Aggregate Functions reuses text from Requirement 31 for Non-spatial functions but this won't work: it needs its own Requirements statement & references to functions like bbox & union in other literature.

I have removed the duplicative text in my "oncethrough" branch but I've not completed the reference citation.

Issue #101, opened 2021-03-02
Spatial Agg Funcs duplicative of existing


nicholascar:
The new _Spatial Aggregate Functions_ are:

* geosaf:BBOX
* geosaf:BoundingCircle
* geosaf:Centroid
* geosaf:ConcatLines
* geosaf:ConcaveHull
* geosaf:ConvexHull
* geosaf:Union

Existing _Non-topological Query Functions_ from 1.0 are:

* geof:distance
* geof:buffer
* geof:convexHull
* geof:intersection
* geof:union
* geof:difference
* geof:symDifference
* geof:envelope
* geof:boundary
* geof:getSRID

1. We have the following duplicates:
    * geosaf:BBOX == geof:envelope
    * geosaf:Union == geof:union
    * geosaf:ConvexHull == geof:convexHull
2. We have introduced a new namespace, `geoaf` when the previous ones used `geof`
3. We have used leading capitals when the previous ones use leading lowercase

Proposal:
* Remove _Spatial Aggregate Functions_ section
* Add missing functions to existing _Non-topological Query Functions_
* Use lower-case letter for function names

So the result will be the following additions to _Non-topological Query Functions_:

1. geosf:boundingCircle
2. geosf:centroid
3. geosf:concatLines
4. geosf:concaveHull

nicholascar:
@situx oops, we should have picked these all up at review time... Do you agree with the proposals above? If not, can you suggest alternates?

situx:
I was aware of these functions, but it was my understanding that we need to define new function signatures in order to state that those can act as spatial aggregate functions. I was not aware that it was simply sufficient to define new non-topological query functions.
Can all defined SPARQL extension functions listed there be used in a SELECT statement, so that we do not need those definitions then or do we need to modify the definitions to state that those may also be used as spatial aggregate functions?

nicholascar:
> it was my understanding that we need to define new function signatures in order to state that those can act as spatial aggregate functions

Sorry, I clearly didn't follow all the thoughts around this - I was doing other PRs...

I think it's confusing having multiple, similar things so I still think we should de-duplicate.

> Can all defined SPARQL extension functions listed there be used in a SELECT statement, so that we do not need those definitions then or do we need to modify the definitions to state that those may also be used as spatial aggregate functions?

Can you supply an example of `geosaf:Union` use in SPARQL v. `geof:union`?

* this may clear things up - i.e. why we need two
[...]

PR #109, opened 2021-03-11
Spatial Aggregate fix

situx:
This pull request works on fixing issues concerning spatial aggregate function definitions referring to #101 and #105

Current issues:

- We have a union function with differing signatures. One is currently renamed to union2. Should we maybe define a function with two different possible signatures?


nicholascar (approved):

nicholascar:
Looks good, although I see you have a merge to deal with!

situx:
Yes I will deal with the merge conflict. However, my concern is still the union2 function. Is that really a good name? I guess we cannot assign two function signatures to one URI, can we?

nicholascar:
OK, just a couple more changes actually:

1.
```turtle
skos:prefLabel "GeoSPARQL Spatial Aggregate Functions Register"
```
remove word "register" as it implies a certain type of vocab (one with statuses for each element):
```turtle
skos:prefLabel "GeoSPARQL Spatial Aggregate Functions"
```

2. Call the function `safuncs:minBoundingCircle` just `safuncs:boundingCircle` as that will better match bounding Box (i.e. we don't call it min bounding box).

3.
```turtle
skos:definition "A spatial aggregate function that calculates a centroid out of a list of vector geometries."@en ;
```
better English:
```turtle
skos:definition "A spatial aggregate function that calculates a centroid of a set of geometries."@en ;
```
and we don't need the qualifier "vector geometries" as we've not used that anywhere else

4.
```turtle
skos:definition "A spatial aggregate function that calculates a concatenated line out of a list of lines."@en ;
```
better English: 
```turtle
skos:definition "A spatial aggregate function that calculates a concatenated line of a set of lines."@en ;
```
5.
```turtle
skos:definition "A spatial aggregate function that calculates the concave hull of a list of vector geometries."@en ;
```
better English:
```turtle
skos:definition "A spatial aggregate function that calculates the concave hull of a set of geometries."@en ;
```
7.
```turtle
skos:definition "A spatial aggregate function that calculates the mininum bounding circle of a given set of geometries."@en ;
```
typo, consistent wording & better English:
```turtle
skos:definition "A spatial aggregate function that calculates the minimum bounding circle of set of geometries."@en ;
```
8.
```turtle
skos:definition "A spatial aggregate function that calculates the union out of a list of vector geometries."@en ;
```
better English:

```turtle
skos:definition "A spatial aggregate function that calculates the union of a set of geometries."@en ;
```

Finally, for the`union2` function name: I actually think `union2` is fine as it's obvious that it is "just" a union function but different from another "just" union function. If you call it something like `aggregateUnion` or `agUnion` or whatever, it looks like it's doing something a bit different, i.e. you can't mistake `2` for indicating any purpose other than identity separation of the function. There are loads of examples in well known software of `function` & `function2` which does the same thing but faster/better etc.

situx:
Ok done

nicholascar:
OK, just waiting for a second review from either @dr-shorthair or @FransKnibbe

dr-shorthair (approved):
Noting, however, that X, Y Z are the typical mathematical notation for 3 dimensions, which may go by other names depending on the CRS (lat/lon/easting/northing/elevation/depth) - does there need to be an informative note added to this effect?

situx:
I think we could add a note for that. In general, I would assume that e.g. by X we mean the first dimension irrespective of the axis order defined by the CRS.

PR #122, opened 2021-03-28
RDF examples

nicholascar:
Examples of the use of each ontology element.

This will stay as a draft PR until all elements have examples.

This example content will need good reviews from several editors for, though it's non-normative, we all know that people will often copy examples quite closely.

See the Examples (Annex B) rendered here: https://raw.githack.com/opengeospatial/ogc-geosparql/rdf-examples/1.1/spec/rdf-egs.html#_annex_b_informative_geosparql_examples

PR #124, opened 2021-03-30
Add example anchors and links

FransKnibbe:
This PR adds ID attributes to examples in the documentation. These ID's can be used in the ontology to form skos:example links.


nicholascar:
Thanks @FransKnibbe, I'll merge this in directly - since it's just documentation enhancement - and then I'll absorb these changes into the PR #122 where I will follow your lead and create IDs for all the new examples and link normative text sections to them.

nicholascar (approved):

PR #125, opened 2021-03-30
Add property :hasDefaultGeometry to geo.ttl

### LINKED ISSUE #123: geo:hasDefaultGeometry is missing in the ontology
Strange: property `geo:hasDefaultGeometry` is described in the specification, but it seems to be missing in geo.ttl in version 1.0. And consequentially it is missing in 1.1. It is also absent from the currently published ontology (http://www.opengis.net/ont/geosparql).

I can add the property if someone can confirm this is a bug.

FransKnibbe:
Add property :hasDefaultGeometry to geo.ttl in version 1.1.  It is still absent in version 1.0 (assuming it won't be published again).
Fixes #123


mperry455 (approved):

nicholascar (approved):

nicholascar:
Thanks again @FransKnibbe: I have stylized the prefLabel the same as other prefLabels (all lower case).

Issue #130, opened 2021-04-09
Is there a stable URI to refer to GeoSPARQL drafts?


andrea-perego:
I would like to report an issue concerning existing specifications which are already citing the new version of GeoSPARQL under preparation.

As far as I can see, the URIs of the current drafts are those linked to from

https://opengeospatial.github.io/ogc-geosparql/

What is not clear is if these URIs are supposed to be stable over time, and can be safely included in bib references when citing the relevant GeoSPARQL drafts / versions.

A related issue is that the HTML rendering of the draft does not carry any date - see, e.g.:

https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html

Moreover, there is no one-to-one (intuitive) match between the relevant build and the content of the GitHub repo. 

As a result, it is not easy to check if the HTML version of the draft is aligned with the latest changes in the AsciiDoc source files. 

Thanks!

situx:
We use a continuous integration process to build the HTML documentation whenever changes are committed to the master branch of this repository (i.e. once pull requests are reviewed and approved by at least two members of the working group)
Therefore, the contents of the HTML version convey the state of the master branch at all times.

It is true that we do not add a date to the current build of the ASCIIDoc and we do not have an automated generation of draft version numbers in the document. The intention was to give anyone interested in the current state of the standard the possibility to get easy access to the standard specification, so we did not really think about giving them a version number as of now.

We did not really talk about whether the drafts should be citable. I would say for now the URIs can be considered stable. Once we publish GeoSPARQL 1.1 we would add a new folder for the GeoSPARQL 1.2 specification and the old URIs should still work unless we restructure the repository, which I think is at least at the moment unlikely.

andrea-perego:
Thanks a lot, @situx .

I'll then consider the current URIs as stable, and always corresponding to the latest version of the living document.

andrea-perego:
@situx , I saw that you plan to have a BibTeX entry via PR https://github.com/opengeospatial/ogc-geosparql/pull/132

An additional option you might consider is to add an entry in SpecRef (https://www.specref.org/), so to make the citation consistently retrievable via a Web API (as done for W3C specs).

In the W3C/OGC Spatial Data on the Web group, we have been already adding entries in SpecRef for a number of OGC standards (GeoSPARQL included -  see https://api.specref.org/bibrefs?refs=geosparql).

situx:
Ok, sure we can consider that as well, thanks for the hint

PR #132, opened 2021-04-15
Citation of the working document

situx:
It appears there is a need for people to already cite the working documents we produce in the continuous integration process, as shown by issue #130 
I think we should give people guidance and define how we want to be cited, by providing a BibTeX definition.
As I am not sure how to best define it I copied a common one from Google Scholar and updated it.
We can modify it until we are satisfied and add it to the Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ To see in-progress HTML & PDF versions of the GeoSPARQL 1.1 standard - specifica
 
 * [GeoSPARQL auto-built documentation](https://opengeospatial.github.io/ogc-geosparql/)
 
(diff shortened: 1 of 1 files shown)


nicholascar (changes requested):
Perhaps just add in the URL: <https://opengeospatial.github.io/ogc-geosparql/>. Always helpful to have that in there!

situx:
Done, shall we consider adding the working document to specref as well?

nicholascar (approved):

nicholascar:
Spec ref would be good!

Issue #112, opened 2021-03-15
Clarification needed for DGGS literal


FransKnibbe:
The GeoSPARQL ontology has been extended with a datatype and property for Discrete Global Grid Systems (DGGS):

    :asDggsWkt
    	a owl:DatatypeProperty ;
    	rdfs:subPropertyOf :hasSerialization;
    	rdfs:domain :Geometry ;
    	rdfs:range :dggsWktLiteral ;
    	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
    	skos:definition """The Discrete Global Grid System (DGGS) serialization of a geometry"""@en ;	
    	skos:prefLabel "as DGGS"@en ;
    .
    :dggsWktLiteral
    	a rdfs:Datatype ;
    	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
    	skos:definition """A textual serialization of a Discrete Global Grid (DGGS) geometry object."""@en ;
    	skos:example """ "<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsWktLiteral>""" ;
    	skos:prefLabel "DGGS Well-Known Text Literal"@en ;
    .

As it stands, I think clarification is needed:

1. The OGC defines a DGGS as a spatial reference system, but GeoSPARQL seems to assume it is a way to describe a geometry. There appears to be a contradiction.
2. There is no (reference to a) definition of a DGGS.
3. There is no reference to a specification of DGGS WKT.

nicholascar:
@FransKnibbe, I'll try to answer as best I can here:

> 1. The OGC defines a DGGS as a spatial reference system, but GeoSPARQL seems to assume it is a way to describe a geometry. There appears to be a contradiction.

I don't think there's a contradiction because each DGGS - which is, indeed, a spatial reference system - also includes within its SRS definition how to represent geometries according to it. So each DGGS will have completely different and incomparable ways of presenting geometries, not like the majority of CRSes which have some form of x/y, long/lat coordinate pairs.

So a geometry for _DGGS X_ will require the use of a literal format defined for that DGGS only.

> 2. There is no (reference to a) definition of a DGGS.

[...]

PR #136, opened 2021-04-22
Update geo.ttl: redefine ontology terms for DGGS

FransKnibbe:
Fixes [issue112](https://github.com/opengeospatial/ogc-geosparql/issues/112) and resolves [issue 137](https://github.com/opengeospatial/ogc-geosparql/issues/137).

### DIFF
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -98,7 +98,7 @@ geosparql:
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
 	skos:definition """A textual serialization of a Discrete Global Grid System (DGGS) geometry object."""@en ;
-	skos:example """<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsLiteral>""" ;
+	skos:example """ "<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsLiteral>""" ;
 	skos:prefLabel "DGGS Literal"@en ;
 .
 
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -98,7 +98,7 @@ geosparql:
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
 	skos:definition """A textual serialization of a Discrete Global Grid (DGGS) geometry object."""@en ;
-	skos:example """ "OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsWktLiteral>""" ;
+	skos:example """ "OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsLiteral>""" ;
 	skos:prefLabel "DGGS Literal"@en ;
 .
 
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -99,7 +99,7 @@ geosparql:
 :dggsLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
-	skos:definition """A textual serialization of a Discrete Global Grid System (DGGS) geometry object."""@en ;
+	skos:definition """A textual serialization of a Discrete Global Grid System (DGGS) object."""@en ;
 	skos:example """ "<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsLiteral>""" ;
 	rdfs:seeAlso <http://www.opengis.net/doc/AS/dggs/2.0> ;
 	skos:prefLabel "DGGS Literal"@en ;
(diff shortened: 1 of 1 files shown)


1.1/geo.ttl
```diff
@@ -94,12 +94,12 @@ geosparql:
...
 .
 
-:dggsWktLiteral
+:dggsLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
 	skos:definition """A textual serialization of a Discrete Global Grid (DGGS) geometry object."""@en ;
-	skos:example """ "<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsWktLiteral>""" ;
-	skos:prefLabel "DGGS Well-Known Text Literal"@en ;
+	skos:example """ "OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsWktLiteral>""" ;
```
dr-shorthair, line 101: Should be `^^<http://www.opengis.net/ont/geosparql#dggsLiteral>`

dr-shorthair (approved):
Fix error in Example, then it is good to go.

1.1/geo.ttl
FransKnibbe: Thank you. I  have just fixed the example.

nicholascar:
I have no problem with this PR in isolation - decomposing a DGGS literal into zones and the DGGS IRI seems fine - but why take this approach when WKT lumps them together? GeoSPARQL 1.0 introduced the extended WKT literal form of:

```
<{SRS-IRI}> {GEOMETRY-TYPE} ((( {COORDINATES} )))
```
 as opposed to just

```
{GEOMETRY-TYPE} ((( {COORDINATES} )))
```

as per the WKT spec.

So why shouldn't we just use DGGS literals like this:

<{DGGS-IRI}> {GEOMETRY-TYPE} ((( {ORDINATES} )))

Is it that you don't like WKT's way of doing things and want to do better this time with a new literal?

FransKnibbe:
Hi @nicholascar,
> So why shouldn't we just use DGGS literals like this:
> 
> <{DGGS-IRI}> {GEOMETRY-TYPE} ((( {ORDINATES} )))
> 
> Is it that you don't like WKT's way of doing things and want to do better this time with a new literal?

That is basically it, yes. See also [issue 31](https://github.com/opengeospatial/ogc-geosparql/issues/31).

1.1/geo.ttl
```diff
@@ -94,12 +94,12 @@ geosparql:
...
 
-:dggsWktLiteral
+:dggsLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
-	skos:definition """A textual serialization of a Discrete Global Grid (DGGS) geometry object."""@en ;
-	skos:example """ "<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsWktLiteral>""" ;
-	skos:prefLabel "DGGS Well-known Text Literal"@en ;
+	skos:definition """A textual serialization of a Discrete Global Grid System (DGGS) geometry object."""@en ;
+	skos:example """<https://w3id.org/dggs/auspix> OrdinateList (R3234)"^^<http://www.opengis.net/ont/geosparql#dggsLiteral>""" ;
```
dr-shorthair, line 101: unbalanced quotes?

nicholascar:
Hi @FransKnibbe, ok I understand your reasoning. Again, I have no issue with this PR as it is and I feel that this decomposition is probably the better way to go (i.e. put distinct properties of geometries into separate RDF elements and only bundle the coordinates/ordinates into a single literal).

I've just spoken with @dr-shorthair and I think we will need to have a wider SWG discussion about this particular issue and also the issue of the `inSRS` property. My hunch is that that property and its implications for WKT and DGGS literals raise difficulties that are beyond the "tidy-up update" planned for GeoSPARQL 1.1 and should be pushed out to GeoSPARQL 1.2, so I may recommend to the SWG that we pull `inSRS` from the current release candidate (1.1). This issue - DGGS - could follow your proposal here - decomposition - but that any implications for existing literals like WKT and Geometries in general would wait for more discussion and for GeoSPARQL 1.2.

1.1/geo.ttl
nicholascar: Sorry, my fault entirely - I altered @FransKnibbe correct work. I've revered my mistake now.

dr-shorthair:
@FransKnibbe can you merge the master into your branch and resolve conflicts?

FransKnibbe:
> @FransKnibbe can you merge the master into your branch and resolve conflicts?

@dr-shorthair: I've tried to merge but I think `:asDGGS` with domain `:Geometry` needs to be `:asDggsLiteral` with domain `:DggsObject`.

jabhay:
Given changes that have taken place since the creation of this ticket and subsequent development in the ticket, some detail has been added to the standard spec to provide a basic handling for DGGS and an example using AusPix. It is proposed that that the existing detail is accepted for 1.1, and that the more detailed handling discussed in this PR be deferred to future releases. Thoughts?

FransKnibbe:
> Given changes that have taken place since the creation of this ticket and subsequent development in the ticket, some detail has been added to the standard spec to provide a basic handling for DGGS and an example using AusPix. It is proposed that that the existing detail is accepted for 1.1, and that the more detailed handling discussed in this PR be deferred to future releases. Thoughts?

Based on earlier discussion at a meeting (I can't recall which exactly), I have removed the separation between a DGGS zones IRI and a literal specifying a spatial object. I do hope that this does not hamper separating them in the future, as is suggested to be done for GeoSPARQL's WKT literal (see[ issue 31](https://github.com/opengeospatial/ogc-geosparql/issues/31)). I was more or less assured in the meeting that copying the structure of `wktLiteral `to `dggsLiteral` would not constitute a precedent in favour of coupling an SRS IRI with a text encoding in a single literal.

The other change, defining a DGGS Object as a subclass of `SpatialObject` instead of a `Geometry` is still much needed, it seems to me. Or have all the issues brought up in [this comment](https://github.com/opengeospatial/ogc-geosparql/issues/112#issuecomment-817356930) been resolved?

PR #140, opened 2021-05-05
Adding Simple Features for SQL Mappings to Annex C

jabhay:
We decided that we wanted to add mappings to other OGC Standards. This Pull Request adds those mappings for Simple Features for SQL.


situx:
Could we create links for the URIs in the table since they should be dereferenceable in the future?

situx (approved):
I added some links and changed the table layout. Could you also add a reference to the Simple features specification document?
Then I believe it would be good to go.

1.1/spec/18-Annex-C.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+The following table maps the functions and properties from Simple Features for SQL to GeoSPARQL.
+
+[frame=none, grid=none, cols="3,3,3,3,3"]
+|===
+| Simple Features for SQL | GeoSPARQL Equivalent | Since GeoSPARQL | Related Property Available | Since GeoSPARQL
+| 2.1.1.1 Basic Methods on Geometry | | | |
+| Dimension(): Double | N/A | - | http://www.opengis.net/ont/geosparql#dimension[geo:dimension] | 1.0
+| GeometryType(): Integer | Class of geometry instance | 1.0 | N/A | -
+| SRID(): Integer | http://www.opengis.net/def/function/geosparql/getSRID[geof:getSRID] | 1.0 | http://www.opengis.net/ont/geosparql#inSRS[geo:inSRS] | 1.1
```
mperry455, line 14: I guess we need to remove inSRS if the group decides to move this property to GeoSPARQL 1.2

mperry455:
There are several simple features functions that could be added as SPARQL functions and/or RDF properties (isEmpty, isSimple, boundary, envelope, centriod, length, area). Right now, some are only functions, some are only properties, and some are both. I wonder if we should do something uniform here and have both a function and a property for each one. The choice of function or property looks kind of arbitrary at the moment.

mperry455 (approved):

situx:
I have removed the inSRS property from the table because we decided in the last meeting not to include it. As the table should show what is implemented until GeoSPARQL 1.1 I would not want to keep speculative references to GeoSPARQL 1.2 in the GeoSPARQL 1.1 standard.
Concerning aligning properties and functions it would make sense to define at least pairs of functions and properties where applicable.

situx:
I added the missing reference and will merge this request for now as we have two approvals. Possibly adding further functions is a separate pull request anyway.

PR #145, opened 2021-05-19
Query example for min and max functions

situx:
I created an example to query min and max coordinates from a polygon geometry


situx:
This request can now be reviewed, as it is. The example for spatial aggregate functions may be put in another pull request

nicholascar (approved):

mperry455 (approved):
Looks good.

PR #146, opened 2021-05-20
Added more functions to conform with Simple Features SQL

situx:
I added more functions that we defined properties for, but which did not have a function representation.

Some functions involve returning a metric value which I normalized to meters in the function specification.

This is probably something to discuss.


nicholascar (approved):
Looks good although we still might have to improve the `dcterms:creator` property here and in all other places (other vocabs) to indicate a Resource rather than a Literal, if we can establish an Agents register

mperry455 (approved):
Looks good.

Issue #126, opened 2021-04-03
Renumbering of requirements for GeoSPARQL 1.1?


situx:
As discussed in #106 we have several requirements that could be grouped in the current specification:
For example:
There are three requirements for every literal type.
We could group them together as requirements 15.1, 15.2, 15.3 instead of using an ascending numbering scheme like in GeoSPARQL 1.0.
What are the best practices to follow from other OGC standards?
How can we align a new numbering system to be consistent with previous work done in GeoSPARQL 1.0?

nicholascar:
> We could group them together as requirements 15.1, 15.2, 15.3 instead of using an ascending numbering scheme like in GeoSPARQL 1.0.

I agree with this: it would be easier to see repeated patters in the Requirements if they were grouped.

> What are the best practices to follow from other OGC standards?

Unfortunately OGC API: Features uses sequential numbering but it is broken own in parts A, B, C etc per number.

> How can we align a new numbering system to be consistent with previous work done in GeoSPARQL 1.0?

This I don’t know. If we are to be backwards-comparable we can’t change everything however the current system of sequential numbers prevents us from adding any new Requirements anywhere other than at the end!

We might just have to implement a new number system and then reference the old numbers in old/new number lookup table in an Annex.

situx:
I think we can discuss this in the next meeting where we in any case wanted to go into a feature freeze, so that we are able to find out about a possible new future-proof numbering theme

mperry455:
I definitely think the a new numbering system with grouped requirements is a good idea. The backwards compatibility part is tricky, but I like the idea of just having a table with the mapping to old requirement numbers somewhere.

FransKnibbe:
I placed a [comment](https://github.com/opengeospatial/ogc-geosparql/pull/106#issuecomment-812913611) on this issue, probably in the wrong place. Long story short: Could putting the requirements in a SKOS concept scheme (and using IRIs to identify them) be a good idea?

PR #152, opened 2021-06-13
Validation Shapes

nicholascar:
This PR is in draft as we test out the appropriateness of including SHACL shapes in 1.1.

SHACL shapes validators are a common element of profiles however no great consideration of the use cases for SHACL shapes or their role v. the ontology rules (i.e. normative, non-normative, itemisation of rules, whether they contribute to conformance classes etc) has been undertaken by the SWG yet.

**A Test Shape**  
`validator.ttl` in this PR contains a single shape to test that `geo:hasGeometry` predicates of `geo:Feature` classes have a `geo:Geometry` range value.

**Questions** 
1. Are shapes like these a good idea?
2. If they are, where do we get all the shape rules from?
    * From the ontology's OWL rules? 
    * What about others?
    * Should shapes be posed for data before or after reasoning?
        * e.g., the example shape will fail data with no `geo:Geometry` class declared as per it's second example, but RDF data post RDFS reasoning will infer that any range value of `geo:hasGeometry` is a `geo:Geometry` and thus the shape will pass
3. How do we link shapes rules/requirements to Abstract Tests, if at all? 
4. What about non-SHACL validators?
    * Does anyone have the interest and motivation to create validators other than SHACL?


nicholascar:
>  include the rules introduced by GeoSHACL

Yes, let's do this.

> I have no experience with other validation techniques apart from SHACL so I would not pursue it. Which other validators would you have in mind?

Neither do I (I've only read about, not used, ShEx) but I just want to leave the call open in case others have strong interests here.

situx:
I have talked to my colleagues creating the GeoSPARQL Compliance benchmark:
Right now we have 
rdfs:isDefinedBy <http://www.opengis.net/ont/geosparql> , <http://www.opengis.net/spec/geosparql/1.0> ;
relations in the GeoSPARQL standard.
How about creating URIs as follows:
<http://www.opengis.net/spec/geosparql/1.0/req/1> 
and so on for all the requirements (which we also represent in RDF)? 
That would also get us more closely to generate a version of the spec from RDF resources only.
Could it not be something to be included in a profile as well?
What we also found is that we often create several tests for a GeoSPARQL requirement. 
For example requirement 19 states:
"Implementations shall support geof:distance, geof:buffer.... etc."
It would make sense to define sub-requirements here for every supported function, maybe <http://www.opengis.net/spec/geosparql/1.0/req/19/1> for the first function to be tested and so on.
Should we do this? I could create a pull request for GeoSPARQL 1.0 in this schema and once the requirement numbers have been finalized for GeoSPARQL 1.1 we could extend upon it.

FransKnibbe:
Hi @nicholascar, here is my response to your questions:
1. yes, I think having a SHACL validator is a very good idea. In a previous project I tried to develop a shapes graph in parallel with a data graph containing meaningful examples. That way the examples are coherent (part of the same story, so to speak) and are continuously syntax and sanity checked. The examples graph is helpful as a test case for the shape graph: you can try to make some mistakes on purpose to see if the validation report picks them up (in the [SHACL playground](https://shacl.org/playground/)). Could this (put all examples in the documentation in the same file/graph, top be validated bu the SHACL file) be a good idea for GeoSPARQL development too?
2. More or less continuing from the first point: the source of the shapes in the first place should be the requirements in the documentation (insofar they can be translated to SHACL), but it could also be our insight into which mistakes people are liable to make.
As for entailment support: it would be nice to have. But is there a SHACL engine that supports inferencing that is free for us (and other GeoSPARQL users) to use?
3. Continuing on the remark by @situx: the idea of having IRIs for requirements[ came up when the numbering of requirements was discussed too](https://github.com/opengeospatial/ogc-geosparql/issues/126#issuecomment-816854922).
4. I think having a single validator should be sufficient. For the user it probably does not matter which validation language is used. One just needs to be able to interpret the validation report.

dr-shorthair:
Formally it is expected to have a URI for each requirement and also for each test, and each requirements-class and each conformance-class. That is part of the OGC modular specification process.

situx:
If that is so which namespace should we use?

nicholascar:
> How about creating URIs as follows:
> http://www.opengis.net/spec/geosparql/1.0/req/1
> and so on for all the requirements (which we also represent in RDF)?
> That would also get us more closely to generate a version of the spec from RDF resources only.

Great suggestion

> Formally it is expected to have a URI for each requirement and also for each test, and each requirements-class and each conformance-class. That is part of the OGC modular specification process.

OK, so we may need a write-up of all the URIs that we need for this 1.1 release, since there are many. Apart from these mentioned here, there are new IRIs for the Profile & the other new profile artifacts, e.g. this SHACL validator.

> If that is so which namespace should we use?

I've started a table for all IRIs that we should request of the OGC NA: https://github.com/opengeospatial/ogc-geosparql/issues/163

nicholascar:
> Could this (put all examples in the documentation in the same file/graph, top be validated bu the SHACL file) be a good idea for GeoSPARQL development too?

I've started "Extended Examples" which are stand-alone resources (i.e. not in the Spec doc) for this sort of thing, see PR #161. There are non-RDF examples there too. We can add deliberately broken examples for SHACL testing.

> As for entailment support: it would be nice to have. But is there a SHACL engine that supports inferencing that is free for us (and other GeoSPARQL users) to use?

Not a SHACL engine (there is) but an OWL engine: [OWL RL](https://github.com/RDFLib/OWL-RL) as well as Java equivalents. SHACL inference is something else, of course, but my point was OWL (or RDFS) intererence. If you want SHACL inference, see [pySHACL](https://github.com/RDFLib/pySHACL).

> I think having a single validator should be sufficient.

OK, let's stick to this for now. Please could you all just directly push to this branch to enhance the validator file?

dr-shorthair:
[OGC-NA Name type specification - specification elements](https://portal.ogc.org/files/?artifact_id=39194)

mathib:
Hi all, I also want to add my two cents here.

1. I believe we should check which SHACL constraints/shapes are useful. TBH, I don't think that the example test shape by Nick is relevant, as Nick points out in the reasoning comment below question 2. The ontology (OWL/RDFS) is descriptive/OWA, while SHACL constraints are restrictive/CWA. In the past, before SHACL existed, OWL/RDFS axioms where "misused" to be prescriptive and I want to avoid this here since (1) there's no need (we can use SHACL now) and (2) it would hopefully lower the confusion people currently have with OWA vs CWA. 
   - To go back to the example of Nick: saying that a node being the target of a `geo:hasGeometry` property should be a `geo:Geometry` instance is not relevant as that information is implicitly already there (the ontology says that all targets of `geo:hasGeometry` are per definition instances of `geo:Geometry`). The only case where this would be required to state, is if you want the users of GeoSPARQL to explicitly assert the classification to `geo:Geometry` in their dataset (but do we really need to require this?) and in that case, the SHACL validation should be done without reasoning of course
   - potentially relevant SHACL constraints IMHO should focus on graph modeling patterns and assisting users with the detection of causes for logical inconsistencies (when an OWL/RDFS reasoner finds an inconsistency, it cannot (should not be able to) decide which part is "erroneous" due to the OWA). The following potentially "relevant" suggestions for SHACL constraints (all open for debate) are partially inspired on a [set of constraints (general and application specific) I created for OMG/FOG/GOM](https://github.com/mathib/shacl_omg-fog) in the past:
      - each node with an incoming `geo:hasGeometry` should have exactly one outgoing relation that is a specialisation of `geo:hasSerialization` (~ geometry nodes without an actual geometry description are not allowed)
      - each node with an outgoing relation that is a specialisation of `geo:hasSerialization`, should have exactly one incoming `geo:hasGeometry` relation (~ geometries without connection to a FoI are not allowed)
      - a node that has an outgoing `geo:hasGeometry` property cannot have an incoming `geo:hasGeometry` property at the same time (a reasoner can detect the inconsistency (a `geo:Feature` cannot be a `geo:Geometry` at the same time), but it cannot be used to pinpoint the problem as SHACL can)
      - checks on the target of datatype properties, i.e. are they effectively RDF literals? (a reasoner can only detect inconsistencies)
      - checks on the datatype of datatype properties, e.g. a `geo:asGeoJSON` property needs to point to a literal with datatype `geo:geoJSONLiteral` (a reasoner cannot infer a datatype via `rdfs:range`, only detect inconsistencies)
   - We can go far here, but I believe we should mainly address the basic modeling patterns implied by the GeoSPARQL spec (and I think there's a lot of open issues). End users can add more specific constraints to these base constraints for their specific situation. 
   - we should also consider if the SHACL constraints are normative or not (or maybe we can split them in a normative part and a non-normative part)
2. Considering the above, I think we can only derive a portion of relevant constraints from the OWL ontology as that was/is meant to be descriptive instead of prescriptive. Expected GeoSPARQL graph modeling patterns is something we should consider as a group, i.e. what graph modeling patterns are suitable when applying the concepts from the GeoSPARQL ontology in a dataset, and I'm afraid we'll have to do this "manually" (i.e. they have not yet been defined) as the result of group discussions.
   - for using SHACL constraints in combination with reasoning: I believe the SHACL spec defines a reasoning flag to indicate that a certain SHACL shape should be tested without or with reasoning (not 100% sure though if this is the intention and if it's widely supported).
3. I would propose to keep the SHACL constraints in a separate document, but linked to the GeoSPARQL profiles and/or GeoSPARQL ontology. Links to requirements would be a nice to have
4. no specific requirements from my side for non-SHACL (e.g. SheX) validators, mainly because they are not standardized and thus less widespread compared to SHACL validation tools...



P.S. The SHACL spec defines how different parts of SHACL can be used for defining constraints vs inference rules (similar to RDFS, OWL, SWRL but allowing more freedom since it's based on SPARQL). I believe this post (by the title) is on the SHACL constraints part and how it can/should be combined with any kind of reasoning but most likely RDFS/OWL since that's the language we used in the GeoSPARQL ontology definition

nicholascar:
@mathib these are excellent suggestions! The OMG/FOG/GOM tests you give examples of are, I think, far better SHACL test than my example one for the reasons you explain. A short series of shapes such as the ones you indicate could, as you say, "mainly address the basic modeling patterns implied by the GeoSPARQL spec".

Would you be willing to contribute your shapes to this PR?

> we should also consider if the SHACL constraints are normative or not 

I think that more helpful than a normative/non-normative split (the traditional standards way of wording things) we can instead just offer a "validation" resource alongside the "specification", which is the Profiles Vocabulary terminology coming through. This better caters for related but different things and even solves the issue of document specification v. OWL ontology (the first is a "specification" the second a "schema" that are related but neither superior to the other). This way, the "validation" resource (there can be many, in different forms) can be used by people who want to validate things and our job - standards editors - is to just ensure maximum consistency between the ontology, specification and validators.

> Considering the above, I think we can only derive a portion of relevant constraints from the OWL ontology 

Agree.

> I would propose to keep the SHACL constraints in a separate document, but linked to the GeoSPARQL profiles and/or GeoSPARQL ontology. Links to requirements would be a nice to have

Agreed and I propose the Profiles Vocabulary mechanism for this, as described above.

---

So, it sounds to me like you have good starting point SHACL tests and we have a common approach to presenting the resource in the standard. Therefore, can you please contribute some useful shapes to this PR (just push directly to the branch) and then, when we have a few, actually useful, shapes, we can confirm the presentation. At that point we might then, as a SWG, discuss what total set of shapes we might aim for. I think just a few would be good but I understand this is a judgement call.

mathib:
> Would you be willing to contribute your shapes to this PR?

Sure! I'll need some time though to make the first attempt. I believe that certain constraints will start discussions, but I see this as a good thing as it will help to define the base modelling patterns in a uniform, explicit and checkable way.

> At that point we might then, as a SWG, discuss what total set of shapes we might aim for. I think just a few would be good but I understand this is a judgement call.

This all depends on how strict we want to make the standard. The stricter, the better the reuse of data will be (queries follow the base modeling patterns) but some modeling freedom might get lost. It's a case by case discussion that we'll need to have I think.

I'll try to make some time for this on Monday, but cannot promise anything :) I noticed that I also had to mention cardinality constraints (e.g. a geometry node can have max 1 accuracy connected)

nicholascar:
> I'll try to make some time for this on Monday...

OK, fingers crossed. I'll perhaps put in some of the form you suggested unless you get to them first.

1.1/validator.ttl
```diff
@@ -21,22 +21,31 @@ As of GeoSPARQL 1.1, this validator is not normaltive, only informative, however
...
+	dcterms:modified "2021-06-22"^^xsd:date ;
 	owl:versionInfo "OGC GeoSPARQL 1.1" ;
 	owl:versionIRI <http://www.opengis.net/def/geosparql/validator/1.1> ;  # same as Spec version for now
 .
 
-<feature-hasGeometry>
+<hasGeometry-hasSerialization>
     a sh:NodeShape ;
-	sh:targetClass geo:Feature ;
+	sh:targetSubjectsOf geo:hasGeometry ;
```
mathib, line 31: targetObjectsOf

1.1/validator.ttl
```diff
@@ -21,22 +21,31 @@ As of GeoSPARQL 1.1, this validator is not normaltive, only informative, however
...
+			sh:alternativePath (
+				geo:hasSerialization
+				geo:asGML
+				geo:asWKT
+				geo:asGeoJSON
+				geo:asKML
+				geo:asDGGS
+			) 
+		] ;
+        sh:minCount 1 ;
```
mathib, line 43: personally, I would also add a `sh:maxCount 1`, i.e. each geo:Geometry has only one geometry serialization. Adding another geometry serialization to the same node, might result in wrong assertions (e.g. metadata concerning CRS, accuracy, etc). Making another serialization often also introduces rounding errors and other small changes that make that it's not 100% the same geometry anymore. TBD

1.1/validator.ttl
```diff
@@ -0,0 +1,154 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@base <http://www.opengis.net/def/geosparql/validator/rule/> .
+
+
+<http://www.opengis.net/def/geosparql/validator>
+	a owl:Ontology ;
+  	dcterms:title "GeoSPARQL RDF Shapes Validator"@en ;
+	dcterms:description """This is a [Shapes Constraint Language (SHACL)](https://www.w3.org/TR/shacl/) file that specifies rules for RDF data. It can be used to test the conformance of data to the GeoSPARQL standard.
+	
+As of GeoSPARQL 1.1, this validator is not normaltive, only informative, however this is likely to be the basis of future, normative, validators."""@en ;
```
mathib, line 16: "normative" (typo)

1.1/profile.ttl
```diff
@@ -13,8 +13,8 @@
     a prof:Profile ;
     owl:versionIRI <http://www.opengis.net/def/geosparql/1.1> ;
     dcterms:created "2020-10-11"^^xsd:date ;
-    dcterms:modified "2020-12-22"^^xsd:date ;
-    dcterms:issued "2021"^^xsd:gYear ;
+    dcterms:creator "OGC GeoSPARQL Standards Working Group" ;
+    dcterms:modified "2021-06-13"^^xsd:date ;
     dcterms:publisher <https://ror.org/00fsdxs93> ;
```
mathib, line 18: let's try to use the same modeling pattern throughout all artefacts

mathib (commented):
just scanned the SHACL file and looks like a good starting point. I've made some small changes.

I'll add the other SHACL from my previous comment this evening or tomorrow (23/06).

We should also test if the SHACL shapes are:
- valid SHACL (https://www.w3.org/TR/shacl/#shacl-shacl)
- are able to track down the issues in the example datasets, incl. transparent reporting

1.1/validator.ttl
nicholascar: >  i.e. each geo:Geometry has only one geometry serialization.

We don't mandate or even suggest that in the Spec and we do have examples of multiple serializations per `Geometry`. [I've shown such examples in Annex B](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b1_1_3_2_a_geometry_with_multiple_serializations) and I have real-world Use Cases that require multiple serilizations, for example, storage of WKT for spatial indexing (Jena) and also as GeoJSON for fast export to web map systems. We even have WKT + DGGS serializations per `Geometry` where the WTK is easy to use with most tooling but the DGGS is super fast for intersection calculations.

> Adding another geometry serialization to the same node, might result in wrong assertions

Maybe, but that's in the camp of user beware! We can't fix people using the wrong CRS etc...

We removed the proposed `inSRS` property (domain `Geometry` so that multiple serializations with differing SRS/CRSes could exist for a single `Geometry`

1.1/validator.ttl
nicholascar: Fixed in Commit https://github.com/opengeospatial/ogc-geosparql/pull/152/commits/747ee942174a35a8f3e2d676b0b4671d4c86fd7f

1.1/validator.ttl
nicholascar: Fixed in Commit https://github.com/opengeospatial/ogc-geosparql/pull/152/commits/747ee942174a35a8f3e2d676b0b4671d4c86fd7f

1.1/profile.ttl
nicholascar: Fixed in Commit https://github.com/opengeospatial/ogc-geosparql/pull/152/commits/747ee942174a35a8f3e2d676b0b4671d4c86fd7f

nicholascar:
> We should also test if the SHACL shapes are:
>
>    * valid SHACL (https://www.w3.org/TR/shacl/#shacl-shacl)
>    * are able to track down the issues in the example datasets, incl. transparent reporting

[pySHACL](https://github.com/RDFLib/pySHACL) reports that the shapes, as they are now, are valid SHACL! 

When I tested against data that deliberately failed the `<hasGeometry-hasSerialization>` test, I got the auto-message from SHACL, _"More than 0 values on data:10->geo:hasGeometry"_, rather than the custom message of _"Each node with an incoming geo:hasGeometry should have at least one outgoing relation that is either geo:hasSerialization or a specialisation of it."_, so some consideration of the best error message reporting needs to be done (i.e. I'm not sure which message is best to report!)

1.1/validator.ttl
mathib: wouldn't it be conceptually cleaner if each `geo:Geometry` nodes has exactly one serialization and geometry nodes of derived geometry serializations are connected? At first sight, this would not seriously harm your perfectly valid use case (use the geometry node with WKT connected for spatial indexing and then traverse the graph by three instead of two steps towards the GeoJSON geometry)? I see direct benefits for provenance (e.g. what is the "original" geometry? What is the chain of derivations?), besides the connection of metadata. If you have both a WKT and GeoJSON with different accuracy/spatialResolution (seems also like a valid use case, no?), how will you document the accuracy/spatialResolution of both with only a single `geo:Geometry` instance (note: `geo:hasSpatialResolution` has `geo:Geometry` as domain)?

It might also be more future proof wrt to later revisions of GeoSPARQL, when CRS/SRS is actually modelled and the GeoJSON is in another CRS/SRS than the WKT geometry, but also when a single geometry consists of a combination of multiple files and thus literals (e.g. Esri Shapes, glTF, OBJ with materials and/or textures, etc), to indicate a certain binary-to-text encoding for binary geometry serializations, for indicating that a geometry is part of another larger geometry (e.g. ref by internal ID), etc. We might postpone this part of the discussion to later, i.e. close the door only when it's needed and leave it open in GeoSPARQL 1.1, but I believe we'll have to do this anyway if we're serious about supporting a broad range of geometry serializations in the future.

FransKnibbe: > If you have both a WKT and GeoJSON with different accuracy/spatialResolution (seems also like a valid use case, no?), how will you document the accuracy/spatialResolution of both with only a single `geo:Geometry` instance (note: `geo:hasSpatialResolution` has `geo:Geometry` as domain)?

That seems like a clear case of multiple `Geometry` instances being needed. Only when geometry serialisations have exactly the same CRS and coordinates there can be multiple serialisations of a single `Geometry`. CRS + coordinates are the defining attributes of a `Geometry`.

FransKnibbe:
> > Could this (put all examples in the documentation in the same file/graph, top be validated bu the SHACL file) be a good idea for GeoSPARQL development too?
> 
> I've started "Extended Examples" which are stand-alone resources (i.e. not in the Spec doc) for this sort of thing, see PR #161. There are non-RDF examples there too. We can add deliberately broken examples for SHACL testing.

Great! I'll start reviewing that PR first then.

mathib:
@nicholascar or someone else: can I get write access to the repository to add some more validation shapes? I've prepared them locally on my machine and tested them with pyshacl. 

I've also provided suggestions for more shapes in the commit, but here we need to discuss first what message we want to give to users:
* only assert specific relations and classes that cannot be inferred from other triples
* leave it open to users to assert potentially redundant information:
   * benefits: no reasoner needed
   * downsides: redundant information needs to be asserted and managed, more validation shapes needed as users can make more errors 

While going through the definition of shapes, I noticed I have some comments on the base ontology, as it's somewhat tied together of course. I'll try to add them tomorrow to existing issues wherever possible.

I would also propose to add a number to the URI of each shape, just to make the testing of the validation shapes a bit easier :). In the current version, I've also named the `sh:PropertyShape`s related to a `sh:NodeShape` and added the `sh:message` on the `sh:PropertyShape` so that it get's emitted correctly (override of the default response produced by the SHACL validation engine).

nicholascar:
> @nicholascar or someone else: can I get write access to the repository to add some more validation shapes?

This is one for @jabhay. If you're not an OGC member, he won't be able to add you however you can still do a PR across forks into this `shapes` branch here. We can then merge those in and then keep this branch under further review and extension before merge to `master`.

jabhay:
Hi @mathib , your access has been granted.

1.1/validator.ttl
nicholascar: > wouldn't it be conceptually cleaner if each geo:Geometry nodes has exactly one serialization and geometry nodes of derived geometry serializations are connected?

Initially I disagreed with the idea of limiting to one serialization per Geometry but I'm coming around to this idea, given the arguments above, especially the provenance one.

So we see Geometry as a class to contain a serialization and other non-serealization elements (e.g. geometry "role" that I've wanted but which isn't present in GeoSPARQL 1.1) but also annotations etc.

OK, if no one else can make any strong objections, I'll convert across to this idea - one serialization per Geometry - enforced with Shapes. I will need to convert examples throughout the Spec & example files to match but I'll only do that after this Shapes PR is finalized. I've raised Issue #172.

FransKnibbe: > OK, if no one else can make any strong objections, I'll convert across to this idea - one serialization per Geometry - enforced with Shapes.

I have to object. I really can´t see any reason why we would want to do that. A good serialisation is able to encode all relevant attributes of a geometry, most importantly its coordinates and CRS. What is wrong with having multiple serialisations that faithfully represent the atttributes of the geometry? Any provenance data that is assigned to a Geometry instance will apply to all serialisations too. Is there something wrong with that? In my opinion, that is the way it is supposed to be.

Look at at this way: We can publish our ontology as RDF/XML, Turtle or JSON-LD. Three different serialisations. But those serialisations do not change the ontology itself. They are just different encodings.

mathib: `Any provenance data that are assigned to a Geometry instance will apply to all serialisations too.` exactly this point is the issue here. This means having exactly the same coordinates, the exact same CRS/SRS, the same binary-to-text encoding, represented accuracy, the same supporting files (textures, material files, etc), etc. 

Even when discussing relatively simple geometry serializations, we notice that geometry formats enforce certain things such as "all geometry in a single fixed coordinate system" (GeoJSON). With future geometry serializations in GeoSPARQL, we get issues of relating additional data to the pure geometry such as textures and/or material information (OBJ requires a specific additional file that's different from what glTF requires, while the latter also allows to combine it with the geometry part). Besides serializations, real world tools (geometry serializers/deserializers) also have their limits. E.g. the same geometry in Rhinoceros CAD exported to OBJ and PLY might result in a slightly different mesh (slightly different vertices, faces, accuracy on coordinates, etc). Sometimes the user has some additional export options to adjust the export, but most often a perfect match is not possible. We need to take this into account and the simplest thing to do is creating different `geo:Geometry` nodes for each serialization and provide provenance links between them if necessary.

I'm also afraid that a lot of users will read past that requirement you mention (all provenance data should apply to all serialisations), resulting in questions or even worse, unreusable datasets.

`We can publish our ontology as RDF/XML, Turtle or JSON-LD. Three different serialisations. But those serialisations do not change the ontology itself. They are just different encodings.` This is only possible because they start from the same data model, i.e. RDF which is profoundly standardized. Geometry descriptions and serializations on the counterpart can vary enormously (e.g. meshes vs NURBS, 2D vs 3D, CAD vs geospatial, etc. resulting in an incredible amount of geometry formats and applications that each are good in a specific set of tasks).

In short, there are in my opinion to many "ifs" for actually making a pattern allowing multiple serializations on a single `geo:Geometry` useful in practice, so that it cannot be widely applied. The other pattern with separate `geo:Geometry` nodes for each serialization adds a tiny bit of complexity but covers way more use cases. Should we therefore allow such a rare pattern, which again increases the possibilities of how something can be modelled (and thus queried) with GeoSPARQL, potentially leading to more confusion and lower adoption?

mathib:
Hi @jabhay , I'm somehow still not able to directly push to the origin/shapes branch from my local repo..

```
ERROR: Permission to opengeospatial/ogc-geosparql.git denied to mathib.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I just followed the workaround of Nick and created a pull request from a fork: https://github.com/opengeospatial/ogc-geosparql/pull/173.

1.1/validator.ttl
FransKnibbe: > `Any provenance data that are assigned to a Geometry instance will apply to all serialisations too.` exactly this point is the issue here. This means having exactly the same coordinates, the exact same CRS/SRS, the same binary-to-text encoding,

Which binary-to-text encoding is that?

> represented accuracy, the same supporting files (textures, material files, etc), etc.

I don't think textures or materials are part of the `Geometry` concept in GeoSPARQL. Geometry is a set of coordinates and the CRS used for those coordinates. Nothing more.


> Even when discussing relatively simple geometry serializations, we notice that geometry formats enforce certain things such as "all geometry in a single fixed coordinate system" (GeoJSON).  With future geometry serializations in GeoSPARQL, we get issues of relating additional data to the pure geometry such as textures and/or material information (OBJ requires a specific additional file that's different from what glTF requires, while the latter also allows to combine it with the geometry part). 

I don't think texture or material are direct attributes of geometry. I suppose additional properties are needed to represent those things, either for the `Geometry` or the `Feature` class (material probably makes most sense as a property of `Feature`).

>Besides serializations, real world tools (geometry serializers/deserializers) also have their limits. E.g. the same geometry in Rhinoceros CAD exported to OBJ and PLY might result in a slightly different mesh (slightly different vertices, faces, accuracy on coordinates, etc). Sometimes the user has some additional export options to adjust the export, but most often a perfect match is not possible. We need to take this into account and the simplest thing to do is creating different `geo:Geometry` nodes for each serialization and provide provenance links between them if necessary.

If the coordinates change because of processing, different `Geometry` nodes are needed. But that is no reason to mandate one serialisation per geometry, I think.

> I'm also afraid that a lot of users will read past that requirement you mention (all provenance data should apply to all serialisations), resulting in questions or even worse, unreusable datasets.

We should do our best to prevent users from wrong GeoSPARQL implementations. That is why I think it is important to have clarified definitions of e.g. "Feature" and "Geometry". If we see a risk of misplacement of provenance data we could mitigate that risk by providing good examples and/or SHACL shapes.


> `We can publish our ontology as RDF/XML, Turtle or JSON-LD. Three different serialisations. But those serialisations do not change the ontology itself. They are just different encodings.` This is only possible because they start from the same data model, i.e. RDF which is profoundly standardized. Geometry descriptions and serializations on the counterpart can vary enormously (e.g. meshes vs NURBS, 2D vs 3D, CAD vs geospatial, etc. resulting in an incredible amount of geometry formats and applications that each are good in a specific set of tasks).

The scope of GeoSPARQL should be knowledge representation, not representation of the wild world of serialisations. In fact, by focusing on the essentials of spatial data, GeoSPARQL could help in reducing all complexity found in formats and applications. 

> In short, there are in my opinion to many "ifs" for actually making a pattern allowing multiple serializations on a single `geo:Geometry` useful in practice, so that it cannot be widely applied. The other pattern with separate `geo:Geometry` nodes for each serialization adds a tiny bit of complexity but covers way more use cases. Should we therefore allow such a rare pattern, which again increases the possibilities of how something can be modelled (and thus queried) with GeoSPARQL, potentially leading to more confusion and lower adoption?

I think the proposed shape is confusing. If a geometry is defined by its coordinates and CRS, why shouldn't you be able to use different serialisations if they have exactly the same coordinates and CRS?

mathib: `Which binary-to-text encoding is that?`
Many options are available (base32, base64, base122, hex, etc) to encode a binary geometry serialization such as binary PLY, GLB, DWG, etc. I don't think we should mandate a single encoding as each of them have their advantages and disadvantages. Leaving the option open to users means that this information should somehow be connectable to the `geo:Geometry` instance, unless you make specific subproperties of a future `geo:asPly-binary` such as `geo:asPly-binary-base32`, `geo:asPly-binary-base64`, etc.

`I don't think textures or materials are part of the Geometry concept`
I certainly agree that (geometry) textures and materials are different from pure geometry, but they are closely related.

`I don't think texture or material are direct attributes of geometry.`
I believe they are. Note that there's a difference between geometry texture and geometry material of a `geo:Geometry` vs physical texture and physical material of a `geo:Feature` described. You might e.g. indicate that the geometry description of a physical chair is yellow to highlight it in a certain visualization, while the actual chair (and thus its physical material) is red.

`The scope of GeoSPARQL should be knowledge representation, not representation of the wild world of serialisations.`
But haven't we been doing this all along GeoSPARQL 1.0 and 1.1 (with good reasons)? If we want to include support for more serializations in GeoSPARQL, should we do a breakdown of the widely used geometry serializations? Don't get me wrong. I do believe there might be a place for a full RDF-based way of describing geometry, e.g. to describe a translation between two geometry serializations or for describing parametric geometry. But I don't think it will ever be as practical and widespread as common, non-RDF geometry serializations (e.g. GeoJSON, OBJ, DWG, glTF, etc.) to be used on a dataset level. Therefore, a practical approach to geometry serializations should also be possible.

FransKnibbe: > > Which binary-to-text encoding is that?
> 
> Many options are available (base32, base64, base122, hex, etc) to encode a binary geometry serialization such as binary PLY, GLB, DWG, etc. I don't think we should mandate a single encoding as each of them have their advantages and disadvantages. Leaving the option open to users means that this information should somehow be connectable to the `geo:Geometry` instance, unless you make specific subproperties of a future `geo:asPly-binary` such as `geo:asPly-binary-base32`, `geo:asPly-binary-base64`, etc.

At the moment we don't have binary serialisations. But if we would have, and we would want to support multiple binary-to-text encodings, those should logically be properties of the serialisation, not of the Geometry.

> > I don't think textures or materials are part of the Geometry concept
> 
> I certainly agree that (geometry) textures and materials are different from pure geometry, but they are closely related.
> 
> > I don't think texture or material are direct attributes of geometry.
> 
> I believe they are. Note that there's a difference between geometry texture and geometry material of a `geo:Geometry` vs physical texture and physical material of a `geo:Feature` described. You might e.g. indicate that the geometry description of a physical chair is yellow to highlight it in a certain visualization, while the actual chair (and thus its physical material) is red.

True. But in GeoSPARQL a geometry is a set of coordinates in a CRS. Not its depiction. How to style or visualise a Geometry is not covered yet, but if it ever was (breaking separation of model and presentation), probably some intermediate construct would be needed.

> 
> > The scope of GeoSPARQL should be knowledge representation, not representation of the wild world of serialisations.
> 
> But haven't we been doing this all along GeoSPARQL 1.0 and 1.1 (with good reasons)?

Well, serialisations are included of course. They are necessary because pure RDF at the moment can be considered inadequate for representing geometries. But GeoSPARQL does not get involved with particularities of serialisations. It refers to the serialisation specifications for that.

If we want to include support for more serializations in GeoSPARQL, should we do a breakdown of the widely used geometry serializations? Don't get me wrong. I do believe there might be a place for a full RDF-based way of describing geometry, e.g. to describe a translation between two geometry serializations or for describing parametric geometry. But I don't think it will ever be as practical and widespread as common, non-RDF geometry serializations (e.g. GeoJSON, OBJ, DWG, glTF, etc.) to be used on a dataset level. Therefore, a practical approach to geometry serializations should also be possible.

Sure. Although I do hope that RDF will evolve to better support for the type of data that Geometry is. But the point remains that Geometry is a clearly defined concept in GeoSPARQL. I think there should be a clear separation between an abstract model for spatial things and ways in which they may be encoded.

nicholascar:
Closing since we have a new Shapes PR

Issue #142, opened 2021-05-06
Proposal: use metre for distance-based properties


FransKnibbe:
Distance-based properties are difficult because of the need to indicate units. This leads to complex expressions of simple numeric data. Moreover, there still are multiple competing ways of expressing units, and there are multiple competing ontologies for units and quantities. 
The problem could be solved by deferring to the distance unit specified in the CRS.  But that poses other difficulties: it requires a lookup, and at this moment we don't have a separate CRS property, nor an agreed-upon ontology or register for CRSs.

A very simple solution is to mimic [hasSpatialResolutionInMeters](https://www.w3.org/TR/vocab-dcat-2/#Property:dataset_spatial_resolution)  as used in DCAT and keep the unit fixed to metre. After all, it is a widely used distance unit and part of the SI. Effectively, only two countries in the world don't use the metre as an official distance unit (source: [wikipedia](https://en.wikipedia.org/wiki/Metric_system#International_System_of_Units)).

This could lead to a set of simple, comparable and interoperable set of properties:

- hasMetricResolution 
- hasMetricAccuracy _(not in GeoSPARQL yet)_
- hasMetricLength
- hasMetricArea _(square metres)_
- hasMetricVolume _(cubic metres)_

 The datatype for metric properties could be `xsd:decimal` (no scientific notation allowed!)
 
 Among the benefits would be:

- Simpler and leaner data publication
- Simpler SPARQL queries
- Better interoperability between properties (e.g. combine resolution, size and accuracy in a single SPARQL query) within a dataset
- Better interoperability between different datasets
- Future proof: no dependencies on external ontologies and developments
- No need for external lookups when processing data
- Less need for numeric conversions in SPARQL queries
- Easier alignment with dataset metadata
- The principle is easy to extend to new properties

 This issue is related to [issue 23](https://github.com/opengeospatial/ogc-geosparql/issues/23
) (spatial accuracy and resolution), [issue 98](https://github.com/opengeospatial/ogc-geosparql/issues/98) (spatial resolution) and [issue 83](https://github.com/opengeospatial/ogc-geosparql/issues/83) (spatial measure).

situx:
I currently wonder how to deal with historical GIS data. What if there is a distance unit we do not yet know how to convert into meters that is mentioned in a historic document?
What if we want to express an ancient unit in our graph because this is the only unit relevant for our target audience?
[...]

Issue #23, opened 2020-09-21
Add concepts for accuracies and tolerances


FransKnibbe:
Contributed by Anna Wagner and Mathias Bonduel for the white paper "Benefits of Representing Spatial Data using Semantic and Graph Technologies":

The AEC domain struggles with geometric representations of planned objects and built objects and corresponding tolerances and inaccuracies, respectively. Planned objects are created on a scratchpad, while construction sites do not offer perfect conditions to recreate the planned geometry completely. Depending on the construction material, the geometry descriptions are commonly enriched with tolerance values, ranging from millimeters (steelwork) to centimeters (masonry). For the geometry of already built objects, the (measured) accuracy is also not perfect, as measuring techniques cannot provide flawless representations. Furthermore, by processing or simplifying geometry descriptions (e.g. from point cloud to mesh), inaccuracies can occur, which are also of interest to attach (represented accuracy). Hence, the possibility to attach accuracies (measured or calculated accuracy) and tolerances would be beneficial for building geometry.

jabhay:
MoSCoW poll created

[![](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Must)](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Must/vote)
[![](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Should)](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Should/vote)
[![](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Could)](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Could/vote)
[![](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Won%27t)](https://api.gh-polls.com/poll/01EJXARVFX9X02KB63YFZZ0AQR/Won%27t/vote)

FransKnibbe:
This issue is about adding ways to express resolution and accuracy. Resolution has been added to GeoSPARQL 1.1 hasSpatialResolution, see also [issue 98](https://github.com/opengeospatial/ogc-geosparql/issues/98). That leaves spatial accuracy. Why don't we add it to GeoSPARQL 1.1 too?

FransKnibbe:
[...]

PR #155, opened 2021-06-14
Add metric properties to the ontology

FransKnibbe:
This adds metric properties to the ontology. It covers [issue 142](https://github.com/opengeospatial/ogc-geosparql/issues/142). 

Note:
- The metric properties do not replace existing properties for size and resolution. That means they can be viewed as convenience properties.
- The PR does include a property for accuracy, which has not been included otherwise yet. I have included it because it was little extra effort.
- ~~If this PR is accepted I can then make a new PR to update the documentation.~~


nicholascar:
Before getting into the specifics of this new content (next comment!), can we have a corresponding update to the Spec doc as well? Every ontology property needs its corresponding Spec write up. Surely somewhere in [Section 8.3](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_standard_properties_for_geogeometry).

nicholascar:
Can the naming of this proposed `hasMetricResolution` be harmonised with the existing `hasSpatialResolution`, i.e. `hasMetricResolution` &rarr; `hasMetricSpatialResolution`?

There seems to be no non-metric equivalent to `hasMetricAccuracy`. Should one be added?

nicholascar:
I am in favour of adding these properties as long as equivalence to the existing non-metric ones are shown, likely just through examples.

FransKnibbe:
> Before getting into the specifics of this new content (next comment!), can we have a corresponding update to the Spec doc as well? Every ontology property needs its corresponding Spec write up. Surely somewhere in [Section 8.3](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_standard_properties_for_geogeometry).
> 
> As a stretch goal: could an example of each of these properties in se be added to [Annex B 1.2](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_2_properties). In particular, could an example show equivalence between direct use of a metric property and the `numericValue` + `unit` examples in [Annex B 1.2.1](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_2_1_feature_properties), just to show that you can either use the metric property or equivalent general property + units of measure.

Ok, I will add both to the PR (I first wanted to be reasonably sure that these additions were acceptable before updating documentation and examples).

FransKnibbe:
> There seems to be no non-metric equivalent to `hasMetricAccuracy`. Should one be added?

I could add it, but to be honest the addition was scheduled for version  1.2 (see [issue 23](https://github.com/opengeospatial/ogc-geosparql/issues/23). But since most to the effort goes into the definition of accuracy, which is already proposed for `hasMetricAccuracy`, I'll add the non-metric equivalent too. It would be strange otherwise. Of course people can still reject adding them both....

FransKnibbe:
I have updated the PR:

- Added hasSpatialAccuracy similar to hasSpatialResolution
- Added documentation
- Added two examples
- Changed datatype from xsd:decimal to xsd:double to allow scientific notation (otherwise I couldn't get the example to work; hurray for examples!)

I think it is ready for re-review now.

nicholascar (changes requested):
Hi @FransKnibbe looks comprehensive with examples, Spec wording so pretty good. I fixed a couple of really tiny typos.

Just one final point from me now: starting with GeoSPARQL 1.1 property naming is of the style **has**XXXX, e.g. _hasGeometry_, but for this set of properties you've not included "has". In keeping with the existing style, e.g. _hasArea_, can you rename your additions with "has", e.g. _hasMetricArea_? Thanks.

FransKnibbe:
> Just one final point from me now: starting with GeoSPARQL 1.1 property naming is of the style **has**XXXX, e.g. _hasGeometry_, but for this set of properties you've not included "has". In keeping with the existing style, e.g. _hasArea_, can you rename your additions with "has", e.g. _hasMetricArea_? Thanks.

@nicholascar, I removed the ¨has¨ on purpose. I got the impression that there are two different styles in use: object properties with ¨has¨ and datatype properties without ¨has¨. It seemed to make some linguistic sense. For example, you could say that a car has an owner and its weight is 1500 kg. The only datatype property with ¨has¨ that I see now is ¨hasSerialisation¨, which kind of confirms the idea that properties starting with ¨has¨ reference more complex things than single values.

But it was just an impression I had, it doesn´t matter much to me either way.

nicholascar:
Looks like not nearly all the ObjectProperties  start with "has", e.g.:

* `ehContains`
* `ehCoveredBy`
* ...
* `sfCrosses`
* ...

My estimate is that "has" was present for `hasGeometry` and we've added `hasArea` etc for the geometry or spatial measure series of properties for a Feature but there was no hard and fast rule for _has_ in GeoSPARQL 1.0.

Clearly we can't alter either the existing `hasGeometre` or the `sf***` & `eh***` properties, so we can't fix on a ObjectProperty (_has_) and a Datatypeproperty (_no has_) approach.

We have _species_ of properties though: 

* spatial relations - `sf***`, `eh***` etc
* Feature/Geometry or SpatialMeasure relations
* datatype properties - `as***` (e.g. `asKML`) and other non-has properties.

In this case of these new properties though, I think they are more similar to the _has_ starting properties they shadow than other datatype properties. i.e. `metricLength` is more similar to `hasLength` than `coordinateDimension` etc, so I'd much prefer all these new properties to start with _has_ too, to match their pairs.

(If I could alter GeoSPARQL 1.0, I would also ad _has_ too all their properties, or remove them all. There is a tend to add _has_ widely (e.g. SOSA & PROF & DCAT) so I'd probably add it, but these options aren't available :( ).

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
 	skos:prefLabel "spatial dimension"@en ;
 .
 
+:metricSpatialResolution
+	a owl:DatatypeProperty ;
+	rdfs:domain :Geometry ;
+	rdfs:range xsd:double ;
```
dr-shorthair, line 595: I prefer `xsd:decimal` - it allows arbitrary precision and is 'honest' in data transfer.

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The spatial resolution of a Geometry in meters."""@en ;
+	skos:note """Spatial resolution specifies the level of detail of a Geometry. It the smallest dinstinguishable distance between spatially adjacent coordinates."""@en;
+	skos:prefLabel "metric spatial resolution"@en ;
+.
+
+:metricSpatialAccuracy
+	a owl:DatatypeProperty ;
+	rdfs:domain :Geometry ;
+	rdfs:range xsd:double ;
```
dr-shorthair, line 605: I prefer `xsd:decimal` - it allows arbitrary precision and is 'honest' in data transfer

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
 .
 
+:metricSpatialResolution
+	a owl:DatatypeProperty ;
+	rdfs:domain :Geometry ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The spatial resolution of a Geometry in meters."""@en ;
+	skos:note """Spatial resolution specifies the level of detail of a Geometry. It the smallest dinstinguishable distance between spatially adjacent coordinates."""@en;
+	skos:prefLabel "metric spatial resolution"@en ;
```
dr-shorthair, line 599: "spatial resolution in meters" - in English at least 'metric' refers to the whole system of units, not just metres.

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+.
+
+:metricSpatialAccuracy
+	a owl:DatatypeProperty ;
+	rdfs:domain :Geometry ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The positional accuracy of the coordinates of a Geometry in meters."""@en ;
+	skos:note """Spatial accuracy is applicable when a Geometry is used to represent a Feature. It is expressed as a distance that indicates the truthfullness of the positions (coordinates) that define the Geometry. In this case accuracy defines a zone surrounding each coordinate within wich the real positions are known to be. The accuracy value defines this zone as a distance from the coordinate(s) in all directions (e.g. a line, a circle or a sphere, depending on spatial dimension)."""@en;
+	skos:prefLabel "metric spatial accuracy"@en ;
```
dr-shorthair, line 609: "spatial accuracy in meters" - in English at least 'metric' refers to the whole system of units, not just metres.

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
 	skos:prefLabel "spatial dimension"@en ;
 .
 
+:metricSpatialResolution
```
dr-shorthair, line 592: This does not follow the DCAT convention where the property name is `*InMeters`. 
However, element IDs are not as important as labels and definitions.

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+	a owl:DatatypeProperty ;
+	rdfs:domain :Geometry ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The spatial resolution of a Geometry in meters."""@en ;
+	skos:note """Spatial resolution specifies the level of detail of a Geometry. It the smallest dinstinguishable distance between spatially adjacent coordinates."""@en;
+	skos:prefLabel "metric spatial resolution"@en ;
+.
+
+:metricSpatialAccuracy
```
dr-shorthair, line 602: This does not follow the DCAT convention where the property name is `*InMeters`. 
However, element IDs are not as important as labels and definitions.

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+	skos:prefLabel "metric spatial accuracy"@en ;
+.
+
+:metricLength
+	a owl:DatatypeProperty ;
+	rdfs:domain :Feature ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The length of a Feature in meters."""@en ;
+	skos:prefLabel "metric length"@en ;
```
dr-shorthair, line 618: "length in meters"

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+.
+
+:metricArea
+	a owl:DatatypeProperty ;
+	rdfs:domain :Feature ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The area of a Feature in square meters."""@en ;
+	skos:prefLabel "metric area"@en ;
```
dr-shorthair, line 627: "area in square meters"

1.1/geo.ttl
```diff
@@ -580,6 +589,53 @@ geosparql:
...
+.
+
+:metricVolume
+	a owl:DatatypeProperty ;
+	rdfs:domain :Feature ;
+	rdfs:range xsd:double ;
+	rdfs:isDefinedBy geosparql: , geosparql-spec: ;
+	skos:definition """The volume of a Feature in cubic meters."""@en ;
+	skos:prefLabel "metric volume"@en ;
```
dr-shorthair, line 636: "volume in cubic meters" - in vernacular English 'metric' denotes the whole system of units, not just metres.

1.1/spec/09-Part-06.adoc
```diff
@@ -191,6 +191,18 @@ TIP: A consistency check can be applied to geometries indicating both this prope
...
+==== Property: geo:metricLength
+The property http://www.opengis.net/ont/geosparql#metricLength[`geo:metricLength`] is similar to http://www.opengis.net/ont/geosparql#hasLength[`geo:hasLength`], but is easier to specify and use because the unit is always meter (the standard length unit of the International System of Units).
+
+```turtle
+geo:metricLength a rdf:Property, owl:DatatypeProperty ;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metric length"@en ;
+    skos:definition "The length of a Feature in meters."@en ; 
+    rdfs:domain geo:Feature ; 
+    rdfs:range  xsd:double .
```
dr-shorthair, line 203: `xsd:decimal` ?

1.1/spec/09-Part-06.adoc
```diff
@@ -217,6 +229,18 @@ TIP: A consistency check can be applied to geometries indicating both this prope
...
+==== Property: geo:metricArea
+The property http://www.opengis.net/ont/geosparql#metricArea[`geo:metricArea`] is similar to http://www.opengis.net/ont/geosparql#hasArea[`geo:hasArea`], but is easier to specify and use because the unit is always square meter (the standard area unit of the International System of Units).
+
+```turtle
+geo:metricArea a rdf:Property, owl:DatatypeProperty ;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metric area"@en ;
+    skos:definition "The area of a Feature in square meters."@en ; 
+    rdfs:domain geo:Feature ; 
+    rdfs:range  xsd:double .
```
dr-shorthair, line 241: `xsd:decimal` ?

1.1/spec/09-Part-06.adoc
```diff
@@ -244,6 +268,18 @@ TIP: A consistency check can be applied to geometries indicating both this prope
...
+==== Property: geo:metricVolume
+The property http://www.opengis.net/ont/geosparql#metricVolume[`geo:metricVolume`] is similar to http://www.opengis.net/ont/geosparql#hasVolume[`geo:hasVolume`], but is easier to specify and use because the unit is always cubic meter (the standard volume unit of the International System of Units).
+
+```turtle
+geo:metricVolume a rdf:Property, owl:DatatypeProperty ;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metric volume"@en ;
+    skos:definition "The volume of a Feature in cubic meters."@en ; 
+    rdfs:domain geo:Feature ; 
+    rdfs:range  xsd:double .
```
dr-shorthair, line 280: `xsd:decimal` ?

1.1/spec/11-Part-08.adoc
```diff
@@ -111,6 +113,46 @@ geo:hasSpatialResolution a rdf:Property, owl:ObjectProperty;
...
+
+The property http://www.opengis.net/ont/geosparql#metricSpatialResolution[`geo:metricSpatialResolution`] is similar to http://www.opengis.net/ont/geosparql#hasSpatialResolution[`geo:hasSpatialResolution`], but it is easier to specify and use because the unit of resolution distance is always meter (the standard distance unit of the International System of Units). 
+
+```turtle
+geo:metricSpatialResolution a rdf:Property, owl:ObjectProperty;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metric spatial resolution"@en ;
+    skos:definition "The spatial resolution of a Geometry in meters."@en ; 
+    rdfs:domain geo:Geometry ;
+    rdfs:range xsd:double .
```
dr-shorthair, line 127: `xsd:decimal` ?

1.1/spec/11-Part-08.adoc
```diff
@@ -111,6 +113,46 @@ geo:hasSpatialResolution a rdf:Property, owl:ObjectProperty;
...
 
+
+==== Property: geo:metricSpatialResolution
+
+The property http://www.opengis.net/ont/geosparql#metricSpatialResolution[`geo:metricSpatialResolution`] is similar to http://www.opengis.net/ont/geosparql#hasSpatialResolution[`geo:hasSpatialResolution`], but it is easier to specify and use because the unit of resolution distance is always meter (the standard distance unit of the International System of Units). 
+
+```turtle
+geo:metricSpatialResolution a rdf:Property, owl:ObjectProperty;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metric spatial resolution"@en ;
```
dr-shorthair, line 124: "spatial resolution in meters"

1.1/spec/11-Part-08.adoc
```diff
@@ -111,6 +113,46 @@ geo:hasSpatialResolution a rdf:Property, owl:ObjectProperty;
...
+```
+
+==== Property: geo:metricSpatialAccuracy
+
+The property http://www.opengis.net/ont/geosparql#metricSpatialAccuracy[`geo:metricSpatialAccuracy`] is similar to http://www.opengis.net/ont/geosparql#hasSpatialAccuracy[`geo:hasSpatialaccuracy`], but it is easier to specify and use because the unit of distance is always meter (the standard distance unit of the International System of Units). 
+
+```turtle
+geo:metricSpatialAccuracy a rdf:Property, owl:ObjectProperty;
+    rdfs:isDefinedBy geo: ;
+    skos:prefLabel "metrci spatial accuracy"@en ;
```
dr-shorthair, line 150: "spatial accuracy in meters"

1.1/spec/11-Part-08.adoc
```diff
@@ -111,6 +113,46 @@ geo:hasSpatialResolution a rdf:Property, owl:ObjectProperty;
...
+    skos:definition "The positional accuracy of the coordinates of a Geometry in meters."@en ; 
+    rdfs:domain geo:Geometry ;
+    rdfs:range xsd:double .
```
dr-shorthair, line 153: `xsd:decimal` ?

1.1/spec/17-Annex-B.adoc
```diff
@@ -76,7 +76,21 @@ Here a `Feature` and a `Geometry` are described as per the previous example but
...
+===== B 1.1.2.4 `Feature` with metric area
+
+
+```turtle
+eg:x a geo:Feature ;
+    skos:prefLabel "Feature X";
+        geo:hasGeometry [
+            geo:asWKT "MULTIPOLYGON (((149.06016596 -35.23610602, 149.060620714 -35.236043434, ... , 149.06016596 -35.23610602)))"^^geo:wktLiteral ;
+    ] ;
+    geo:metricArea 8.7E4^^xsd:double;
```
dr-shorthair, line 88: 87400.

Or `"8.74E4"^^xsd:double`

1.1/spec/17-Annex-B.adoc
```diff
@@ -126,7 +140,25 @@ In this example, `Feature X` has two different `Geometry` instances indicated wi
...
 Machine use of this `Feature` would be able to differentiate the two `Geometry` instnaces based on this use of http://www.opengis.net/ont/geosparql#hasSpatialResolution[`geo:hasSpatialResolution`].
 
-===== B 1.1.2.6 `Feature` with two different types of `Geometry` instances
+===== B 1.1.2.7 `Feature` with two different `Geometry` instances using metric spatial resolution
+
+```turtle
+eg:x a geo:Feature ;
+    skos:prefLabel "Feature X";
+    geo:hasGeometry [
+        geo:metricSpatialResolution 100^^xsd:float ;
```
dr-shorthair, line 149: `100.`

Or `"100"^^xsd:double`

1.1/spec/17-Annex-B.adoc
```diff
@@ -126,7 +140,25 @@ In this example, `Feature X` has two different `Geometry` instances indicated wi
...
+        geo:asWKT "MULTIPOLYGON (((149.06016 -35.23610, 149.060620 -35.236043, ... , 149.06016 -35.23610)))"^^geo:wktLiteral ;
+    ] ,
+    [
+        geo:metricSpatialResolution 5^^xsd:float ;
```
dr-shorthair, line 153: `5.`

Or `"5.0"^^xsd:double`

dr-shorthair:
1. I suggest a bunch of changes 'metric' to 'in meters" because in vernacular English 'metric' denotes the system of units, not just metre. 
2. I prefer `xsd:decimal` because it is more honest in a serialization. But I know there are different opinions on whether the focus should be on data transfer or in-memory representation

FransKnibbe:
@dr-shorthair: thanks for the prompt review.

> 1. I suggest a bunch of changes 'metric' to 'in meters" because in vernacular English 'metric' denotes the system of units, not just metre.

To be clear, you suggest to change only the `skos:prefLabel`s, right? I see not problem with that.

> 2. I prefer `xsd:decimal` because it is more honest in a serialization. But I know there are different opinions on whether the focus should be on data transfer or in-memory representation

I started out with `xsd:decimal`, but then I got into trouble with one of the examples where a value of 8.7 hectometre was used. To express that with  `xsd:decimal` would introduce false precision (8700). So I realized the ability to use scientific notation was needed. I then chose `xsd:double` because it has the highest capability to represent high precision numbers. Floating point numbers have disadvantages, but in my mind they do not outweigh the need for using significant digits.

dr-shorthair:
The words in the label (and textual definition) are the most important. 
However, the property-name also matters - I would suggest also making those `AAAInMeters` rather than `metricAAA`. 

I hadn't thought that a lot of trailing zeros gave an impression of a defined precision. Hmmm. There is another issue with double and float described here https://github.com/qudt/qudt-public-repo/issues/307 Perhaps it is not relevant in this slot, but worth a passing thought at least.

FransKnibbe:
@dr-shorthair: I have made the following changes:

1. Added "has" to the properties
2. Changed the labels from "metric ...." to ".... in meters"

I haven't changed the property names (yet), with the idea that the current names are shorter and sort of form a group because their names all start with "hasMetric"; it makes them a bit easier to distinguish from their counterparts that can have arbitrary units.

Thanks for pointing out the [QUDT issue](https://github.com/qudt/qudt-public-repo/issues/307). It seems to me that the cases are not fully comparable. The QUDT issue is about conversion factors, while the metric properties in GeoSPARQL will be used for values that are measured or estimated. It is possible to imagine a conversion factor that has too many digits for xsd:double. But for GeoSPARQL we could assume that xsd:double offers enough precision to express measured or estimated values. Not being able to prevent false precision would be a much higher penalty, I think.

Can you accept the PR in its current state?

dr-shorthair:
> Thanks for pointing out the QUDT issue. It seems to me that the cases are not fully comparable. 

Yes, I agree that the QUDT issue is addressing a slightly different issue. But thought you would be interested to see a practical analysis of decimal vs. double ;-)

dr-shorthair (approved):

dr-shorthair (approved):

FransKnibbe:
@nicholascar: Could you please take a look at this PR again to see if it can be approved?

nicholascar (approved):

Issue #10, opened 2019-10-18
Make underlying definitions available on the semantic web


lvdbrink:
**Official change request**: [OGC CR 600](http://ogc.standardstracker.org/show_request.cgi?id=600)

GeoSPARQL is currently lacking in semantics that are available on the semantic web. For example, the concept of Geometry is not explained in GeoSPARQL itself, but has in its text definition a pointer to ISO 19107: "This class is equivalent to the UML class GM_Object defined in ISO 19107". Should one look up that document on the web, a paywall from ISO is hit. Such fundamental definitions should be directly accessible to consumers, both human and machine, on the semantic web, either as part of GeoSPARQL or as another web vocabulary.

jabhay:
MoSCoW poll created

[![](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Must)](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Must/vote)
[![](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Should)](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Should/vote)
[![](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Could)](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Could/vote)
[![](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Won%27t)](https://api.gh-polls.com/poll/01EJXB8A9VVT4N9XGDBH9JYYQ1/Won%27t/vote)

jabhay:
@ghobona : It would make sense to me that the underlying infrastructure to accomplish this is already available in the OGC Naming Authority. Would you agree?

If so, my understanding is that we'd elect SWG representatives to the OGC Naming Authority, and then submit a PR into your already existing processes and viola, yes?

FransKnibbe:
[...]

PR #158, opened 2021-06-15
Update 07-Part-04.adoc - terms and definitions

FransKnibbe:
This PR adds terms and definitions for spatial referencing. In the current state, use of the terms CRS and SRS is confusing. The definitions now clearly state that coordinate reference systems are a subset of spatial reference systems,

See also [this comment](https://github.com/opengeospatial/ogc-geosparql/issues/10#issuecomment-825168038) from a  discussion about definition of terms in the ontology.

Notes:

1. I think it would be good to add other terms that are important for understanding GeoSPARQL to section 4 too, as a convenience for the user/reader.
2. If we agree on these definitions. I can go through the ontology and documentation to check for proper usage of CRS and SRS, and make changes where required, in a new PR.

### DIFF
--- a/1.1/spec/07-Part-04.adoc
+++ b/1.1/spec/07-Part-04.adoc
@@ -1,4 +1,11 @@
 == Terms and definitions
 
-For the purposes of this document, the terms and definitions given in the above references apply.
+The terms and definitions given in the above references apply to the GeoSPARQL specification, with the exception of the terms defined below.
 
+**coordinate system:**  A set of mathematical rules for specifying how coordinates are to be assigned to points.
+
+**coordinate reference system:** A **__coordinate system__** that is related to an object by a  **__datum__**.
+
+**datum:** Parameter or set of parameters that define the position of the origin, the scale, and the orientation of a **__coordinate system__**.
+
+**spatial reference system:**  A system for establishing spatial positions. A spatial reference system can use geographic identifiers (e.g. names like "Paris" or "Africa"), coordinates (in which case it is a **__coordinate reference system__**) or identifiers with structured geometry (in which case it is a Discrete Global Grid System (DGGS)). 
--- a/1.1/spec/07-Part-04.adoc
+++ b/1.1/spec/07-Part-04.adoc
@@ -2,10 +2,10 @@
 
 The terms and definitions given in the above references apply to the GeoSPARQL specification, with the exception of the terms defined below.
 
-**coordinate system:**  A set of mathematical rules for specifying how coordinates are to be assigned to points.
+**coordinate system**:  A set of mathematical rules for specifying how coordinates are to be assigned to points.
 
-**coordinate reference system:** A **__coordinate system__** that is related to an object by a  **__datum__**.
+**coordinate reference system**: A **__coordinate system__** that is related to an object by a  **__datum__**.
 
 **datum:** Parameter or set of parameters that define the position of the origin, the scale, and the orientation of a **__coordinate system__**.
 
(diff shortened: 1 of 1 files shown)


nicholascar (changes requested):
@FransKnibbe thanks for this clarification, I think it's needed. I will, as discussed, put in a PR to remove the asCRS property still...

Can I ask for an update though: these terms should either reuse or be linked to terms in the OGC set of vocabs. Searching for "coordinate reference system" on the Defs Server [1] returns some promising results, in particular <https://www.opengis.net/def/CaLAThe/4.0/CoordinateReferenceSystem>. If at all possible we would reuse a term like this or provide a differentiating definition + reference.

If we really do need different definitions (hopefully not!) then we should contribute these terms back.

[1] https://defs.opengis.net/vocprez/search?search=coordinate+reference+system&from=all

FransKnibbe:
> Can I ask for an update though: these terms should either reuse or be linked to terms in the OGC set of vocabs. 

Well, the definitions for coordinate system and coordinate reference system are equal to the definitions in the two OGC documents that are listed as normative for definitions (the GML and SFA specs). The point is (see [the comment in the definitions issue thread](https://github.com/opengeospatial/ogc-geosparql/issues/10#issuecomment-825168038)) that neither spec defines _spatial reference system_.  And in order to have that defined in GeoSPARQL, I felt the closely related terms should be directly defined here too. 

I keep wondering why GeoSPARQL does not have its own set of definitions. Most other OGC standards have glossaries of terms, see for example the two normative specs. Having a glossary within the specification is much more user-friendly than pointing to a set of nine documents where the definition of an unknown term may be found. And I think a user-friendly lookup of definitions   is especially important  for GeoSPARQL because people from multiple backgrounds will try to follow the specification. Many GIS people will be unfamiliar with RDF terms and many (sem)web people will be unfamiliar with terms from geoinformatics.

So actually having only these four definitions is half-hearted. Why not have a full glossary, like other OGC specifications?

> Searching for "coordinate reference system" on the Defs Server [1] returns some promising results, in particular https://www.opengis.net/def/CaLAThe/4.0/CoordinateReferenceSystem. If at all possible we would reuse a term like this or provide a differentiating definition + reference.

> If we really do need different definitions (hopefully not!) then we should contribute these terms back.
> 
> [1] https://defs.opengis.net/vocprez/search?search=coordinate+reference+system&from=all

The Defs Server could be a good replacement of glossaries in specification documents. But it does not seem ready for that yet. Searching for e.g. "coordinate reference system" results in a long list of documents, instead of a single common definition.

nicholascar:
> I keep wondering why GeoSPARQL does not have its own set of definitions. 
> So actually having only these four definitions is half-hearted. Why not have a full glossary, like other OGC specifications?

I presume just an oversight? Perhaps since this standard defines an Ontology, there was less emphasis placed on definitions in the Specification document?

> The Defs Server could be a good replacement of glossaries in specification documents. But it does not seem ready for that yet. 

I agree. Perhaps we can make a full glossary and just ensure that each term defined then links back to any OGC NA or others, such as W3C, point-of-truth definitions. This will mean the glossary is easy to read but tied to the other definition systems.

---

OK, would you like to propose a full term set for the glossary? Perhaps you could add a table in a comment here and others can add to it. I presume you'd want RDF/OWL things like:

* Resource
* Subject
* Predicate
* Object
* Class
* Property
* Object Property
* Datatype Property
* Annotation Property

situx:
I would also think a glossary is a very useful thing to have in the spec

FransKnibbe:
Thank you for your thoughts, @nicholascar  and @situx. I have made a new issue for making a full glossary ([issue 164](https://github.com/opengeospatial/ogc-geosparql/issues/164)) and assigned it to myself. I will withdraw this PR, which is to be replaced by a comprehensive one.

FransKnibbe:
This PR will be replaced by a PR for a complete glossary of terms

PR #161, opened 2021-06-16
Extended examples

nicholascar:
Examples that are too long for the Spec Annex. Please note that the examples are considered another profile resource and thus are listed in the profile declaration.


dr-shorthair (commented):
The examples are not consistent. .auspix and .gml are pure geometry. 
The others all have the geometry nested under some other structure. Could they be made more uniform?

nicholascar:
> The examples are not consistent.

I agree. We have confined the equivalent RDF literals to just the geometries but I thought I'd supply full Feature instances here. Problem is, there is no standard way of describing an AusPIX/DGGS Feature, other than... GeoSPARQL 1.1!

I can fix the GML one, but AusPIX might just have to stay like that.

nicholascar:
@dr-shorthair @FransKnibbe @situx can you please all review so we can merge this in. The intent is to extend the examples perhapps continuously so please allow this PR in then add to the examples. This PR will tidy up some of the files in the repo too (e.g. removing old rdf-egs.html docs)

situx (approved):
Looks ok to me

FransKnibbe:
Hi @nicholascar. it looks good at a glance (and knowing it is a work in progress).
Two questions though:

1. Why use the ^^xsd:anyURI notation instead of (and next to) turtle's <IRI> notation? I would opt for using <IRI> always.
2. locn is one of the default namespaces. Why not use locn:geographicName instead of dcterms:title for geographical names?

nicholascar:
> Why use the ^^xsd:anyURI notation instead of (and next to) turtle's <IRI> notation? I would opt for using <IRI> always.
    locn is one of the default namespaces. 

`<IRI>` and `xsd:anyURI`  are for completely different things:
* `<IRI>` indicates the node is an RDF resource - the IRI acts as the identifier and, in the best case, the IRI can be dereferenced to get more RDF thus it may extend the Linked Data web
* `xsd:anyURI` is a datatype specialised for a URL, so the thing it applies to is a literal, not a resource, and while it's a URL, as opposed to a number, generic string, token, etc., it's not expected to be able to be dereferenced to get more RDF. It's not extending the Linked Data web


e.g. `rdfs:seeAlso "https://github.com/GeoscienceAustralia/AusPIX_DGGS"^^xsd:anyURI ;` - here the object is a literal - a URL, not a resource. We can expect a person to "look up" this web address (URL) as opposed to a machine using it as an ID for an RDF node and perhaps trying to resolve it for more RDF. This URL, in GitHub, can never return RDF.


> locn is one of the default namespaces

Default for what, a particular system you are using? It's hardly widely used and I only come across it in DCAT things and I'm hoping to see it phased out in favour of GeoSPARQL, now that GeoSPARQL has things that DCAT requires, like Bounding Box, Centroid etc.

> Why not use locn:geographicName instead of dcterms:title for geographical names?

1. Because I make no use of locn
2. I use dcterm:title widely (for Datasets, for FeatureCollections etc.) so it's just more use of something widely known to continue using it

FransKnibbe:
> > locn is one of the default namespaces
> 
> Default for what, a particular system you are using? It's hardly widely used and I only come across it in DCAT things and I'm hoping to see it phased out in favour of GeoSPARQL, now that GeoSPARQL has things that DCAT requires, like Bounding Box, Centroid etc.

Actually, locn is a prefix defined in demo-dataset.ttl. Your own file :-). That is what I meant by "default namespace".

I think of locn as an existing ontology for spatial referencing by identifiers. The upcoming terms & definitions will mention three types of spatial referencing: by coordinates (CRS), by predefined structured geometry (DGGS) and by identifiers (e.g. geographic names). With CRS and DGGS covered by GeoSPARQL, LOCN could be considered complementary to GeoSPARQL.

nicholascar:
> Actually, locn is a prefix defined in demo-dataset.ttl. Your own file :-).

I've been caught out! So it's not true that "I make no use of locn", but I certainly don't use it much! Oh, well, I don't think the single use of it for `locn:geometry` is preblematic: this extended example is showing locn side-by-side with GeoSPARQL for geometry lining, so that's useful.

PR #165, opened 2021-06-22
Terms & Defs vocab

nicholascar:
This is a start to a Terms & Defs vocab in SKOS that will be auto-converted to ASCIIDOC for Clause 4 in the Specification though an enhancement to pyLODE.

I have defined a few RDF concepts and placed them into an "Semanitc Web" Collection within the ConceptScheme to group them nicely.

These are really difficult to define due to the unhelpful nature of RDF/RDFS/OWL definitions like, for `rdfClass` _"The class resource, everything."_. I've faithfully reproduced the original definitions abut then added what I think we actually want people to read in a `scopteNote`. 

@dr-shorthair what do you make of this pattern of original defn -> `skos:definition` + useful defn -> `skos:skopeNote`? Should we instead just add the other useful den to a single `skos:definition`? 

I've also indicated that some of the SKOS Concepts here are `owl:equivalentClass` to the originally defined things. This may be unnecessary: such relations will only be potentially relevant for RDF/OWL things and not for non Semantic Web definitions, e.g. spatial things.

@rob-metalinkage please could you also review to see what the OGC NA might make of vocabs like these?

### DIFF
--- a/1.1/terms_and_defns.ttl
+++ b/1.1/terms_and_defns.ttl
@@ -1,0 +1,159 @@
+@prefix cs: <http://www.opengis.net/def/geosparql/terms-and-definitions/> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix prof: <http://www.w3.org/ns/dx/prof/> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sdo: <https://schema.org/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix td: <http://www.opengis.net/def/geosparql/terms-and-definitions/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+cs:
+    a owl:Ontology , skos:ConceptScheme ;
+    dcterms:created "2021-06-22"^^xsd:date ;
+	dcterms:creator "OGC GeoSPARQL Standards Working Group" ;
+    dcterms:modified "2021-06-22"^^xsd:date ;
+    dcterms:provenance "Each term (Concept) listed here indicates its own provenance since they are derived from other vocabularies and non-vocabulary sources."@en ;
+	dcterms:publisher [
+		a sdo:Organization ;
+		sdo:name "Open Geospatial Consortium" ;
+		sdo:url "https://www.ogc.org"^^xsd:anyURI ;
+	] ;
+    owl:versionInfo "OGC GeoSPARQL 1.1" ;
+    skos:definition "A vocabulary (taxonomy) of the Terms and Definitions listed by the GeoSPARQL 1.1 standard. These terms appear as a vocabulary resource (this file) and are also listed in Clause 4 of the GeoSPARQL 1.1 Specification"@en ;
+    skos:hasTopConcept
+        td:Subject ,
+        td:Predicate ,
+        td:Object ,
+        td:Resource ,
+        td:Class ,
+        td:Property ,
+        td:ObjectProperty ;
+        # td:DatatypeProperty ,
+        # td:AnnotationProperty 
+    skos:prefLabel "GeoSPARQL Terms & Definitions" ;
+.
+
+# Concept Template
+# td:XXX a skos:Concept ;
+#     dcterms:provenance ""@en ;
+#     rdfs:isDefinedBy <> ;
+#     skos:definition ""@en ;
+#     skos:inScheme cs: ;
+#     skos:prefLabel ""@en ;
+#     skos:topConceptOf cs: ;
+# .
+
+
+td:SemanticWeb a skos:Collection ;
+    dcterms:provenance "This Colelction was created for GeoSPARQL 1.1's Specification's Terms & Definitions Clause"@en ;
+    skos:definition "Concepts from the Semantic Web domain"@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "Semantic Web Concepts"@en ;
+    skos:member        
+        td:Subject ,
+        td:Predicate ,
+        td:Object ,
+        td:Resource ,
+        td:Class ,
+        td:Property ,
+        td:ObjectProperty ;
+        # td:DatatypeProperty ,
+        # td:AnnotationProperty 
+.
+
+td:Subject a skos:Concept ;
+    dcterms:provenance "From the RDF Concepts Vocabulary (RDF)"@en ;
+    rdfs:isDefinedBy rdf: ;
+    skos:definition "The subject of the subject RDF statement."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDF Subject"@en ;
+    skos:scopeNote """The RDF Concepts Vocabulary (RDF)'s models all connected ideas through the use of triples - 3-part association - made of subject, predicated & objects. This term indicate the subject of a triple.
+    
+In this set of terms, this object is names 'RDF Subject' to distinguish it from non-RDF resources, given that the word 'subject' is general in nature."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:Predicate a skos:Concept ;
+    dcterms:provenance "From the RDF Concepts Vocabulary (RDF)"@en ;
+    rdfs:isDefinedBy rdf: ;
+    skos:definition "The predicate of the subject RDF statement."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDF Predicate"@en ;
+    skos:scopeNote """The RDF Concepts Vocabulary (RDF)'s models all connected ideas through the use of triples - 3-part association - made of subject, predicated & objects. This term indicate the predicate of a triple.
+    
+In this set of terms, this object is names 'RDF Predicate' to distinguish it from non-RDF resources."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:Object a skos:Concept ;
+    dcterms:provenance "From the RDF Concepts Vocabulary (RDF)"@en ;
+    rdfs:isDefinedBy <http://www.w3.org/2000/01/rdf-schema#> ;
+    skos:definition "The object of the subject RDF statement."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDF Object"@en ;
+    skos:scopeNote """The RDF Concepts Vocabulary (RDF)'s models all connected ideas through the use of triples - 3-part association - made of subject, predicated & objects. This term indicate the object of a triple.
+    
+In this set of terms, this object is names 'RDF Object' to distinguish it from non-RDF resources, given that the word 'object' is general in nature."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:Resource a skos:Concept ;
+    dcterms:provenance "From the RDF Schema vocabulary (RDFS)"@en ;
+    owl:equivalentClass rdfs:Resource ;
+    rdfs:isDefinedBy rdfs: ;
+    skos:definition "The class resource, everything."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDFS Resource"@en ;
+    skos:scopeNote """The RDF Schema vocabulary (RDFS)'s 'Resource' is the most general RDF element and all things defined by RDF are some specialization of 'Resource'. 
+    
+In this set of terms, this object is names 'RDFS Resource' to distinguish it from non-RDF resources, given that the word 'resource' is general in nature."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:Class a skos:Concept ;
+    dcterms:provenance "From the RDF Schema vocabulary (RDFS)"@en ;
+    owl:equivalentClass rdfs:Class ;
+    rdfs:isDefinedBy rdfs: ;
+    skos:definition "The class resource, everything."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDFS Class"@en ;
+    skos:scopeNote """The RDF Schema vocabulary (RDFS)'s 'Class' is set notion, i.e. a type of thing for which there are a number of individuals of that type (there could be zero individuals). For example, the class of 'Hammer' is a set of individuals - individual hammers - with certain properties that are defined to be the essential properties of a hammer.
+    
+In this set of terms, this object is names 'RDFS Class' to distinguish it from non-RDF resources, given that the word 'class' is general in nature."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:Property a skos:Concept ;
+    dcterms:provenance "From the RDF Schema vocabulary (RDFS)"@en ;
+    owl:equivalentClass rdfs:Property ;
+    rdfs:isDefinedBy rdfs: ;
+    skos:definition "The class of RDF properties."@en ;
+    skos:inScheme cs: ;
+    skos:prefLabel "RDFS Property"@en ;
+    skos:scopeNote """The RDF Schema vocabulary (RDFS)'s 'Property' is the class of all types of things that can be properties of an individual of any class.
+    
+In this set of terms, this object is names 'RDFS Property' to distinguish it from non-RDF resources, given that the word 'class' is general in nature."""@en ;
+    skos:example """owl:ObjectProperty is a special type of rdf:Property that links the subject of an RDF triple to the object of that triple where the object is not a simple data type."""@en ;
+    skos:example """owl:DatatypeProperty is a special type of rdf:Property that links the subject of an RDF triple to the object of that triple where the object is a simple data type, such as a literal text value."""@en ;
+    skos:topConceptOf cs: ;
+.
+
+td:ObjectProperty a skos:Concept ;
+    dcterms:provenance "The OWL 2 Schema vocabulary (OWL 2)"@en ;
+    owl:equivalentClass owl:ObjectProperty ;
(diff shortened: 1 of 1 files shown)


dr-shorthair:
`skos:definition` should hold a single precise definition. `skos:scopeNote` should hold a gloss. If you want to provide an alternative phrasing of the definition, probably best to use a generic `skos:note`. 

The [ISO/IEC directives](https://www.iso.org/sites/directives/current/part2/index.xhtml) has guidance about what constitutes a good definition - see [clause 16.5](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor218) in particular 16.5.6: 
- The definition shall be written in such a form that it can replace the term in its context. It shall not start with an article (“the”, “a”) nor end with a full stop. A definition shall not take the form of, or contain, a requirement.
- Only one definition per terminological entry is allowed. If a term is used to define more than one concept, a separate terminological entry shall be created for each concept and the domain shall be included in angle brackets before the definition.
- Circular definitions, which repeat the term being defined, are not allowed.

Also note that the ISO/IEC directives encourage examples - use `skos:example` of course.

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+        td:Property ,
+        td:ObjectProperty ;
+        # td:DatatypeProperty ,
+        # td:AnnotationProperty 
+    skos:prefLabel "GeoSPARQL Terms & Definitions" ;
+.
+
+# Concept Template
+# td:XXX a skos:Concept ;
+#     dcterms:provenance ""@en ;
```
FransKnibbe, line 41: The way `dcterms:provenance` is used now in the concept scheme seems to have more to do with the source. So let's use `dcterms:source` for that? And use `dcterms:provenance` for documenting changes (which at the moment is not required because the concept scheme is fresh).

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sdo: <https://schema.org/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix td: <http://www.opengis.net/def/geosparql/td/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+cs:
+    a owl:Ontology , skos:ConceptScheme ;
```
FransKnibbe, line 14: Why is this an `owl:Ontology` too?

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+    dcterms:modified "2021-06-22"^^xsd:date ;
+    dcterms:provenance "Each term (Concept) listed here indicates its own provenance since they are derived from other vocabularies and non-vocabulary sources."@en ;
+	dcterms:publisher [
+		a sdo:Organization ;
+		sdo:name "Open Geospatial Consortium" ;
+		sdo:url "https://www.ogc.org"^^xsd:anyURI ;
+	] ;
+    owl:versionInfo "OGC GeoSPARQL 1.1" ;
+    skos:definition "A vocabulary (taxonomy) of the Terms and Definitions listed by the GeoSPARQL 1.1 standard. These terms appear as a vocabulary resource (this file) and are also listed in Clause 4 of the GeoSPARQL 1.1 Specification"@en ;
+    skos:hasTopConcept
```
FransKnibbe, line 26: I'd like to suggest having two top concepts, one for the semantic web and one for geoinformatics/OGC

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+#     dcterms:provenance ""@en ;
+#     rdfs:isDefinedBy <> ;
+#     skos:definition ""@en ;
+#     skos:inScheme cs: ;
+#     skos:prefLabel ""@en ;
+#     skos:topConceptOf cs: ;
+# .
+
+
+td:SemanticWeb a skos:Collection ;
```
FransKnibbe, line 50: Is a skos:Collection the right thing to use here? It seems intended for array-like structures. Shouldn't we use broader/narrower relationship to form a hierarchy?

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+.
+
+# Concept Template
+# td:XXX a skos:Concept ;
+#     dcterms:provenance ""@en ;
+#     rdfs:isDefinedBy <> ;
+#     skos:definition ""@en ;
+#     skos:inScheme cs: ;
+#     skos:prefLabel ""@en ;
+#     skos:topConceptOf cs: ;
```
FransKnibbe, line 46: So all terms are to be top concepts? I suggest changing this to skos:broader, to link to one of the top concepts (semweb terms and GIS terms).

FransKnibbe:
Hi @nicholascar, I have mainly looked at the file as a template before new terms and definitions are added. My main point that if we define semweb and GIS terms as the two top concepts the concept scheme seems to be more in line with the intentions described in the [SKOS primer](https://www.w3.org/TR/skos-primer/).

1.1/terms_and_defns.ttl
nicholascar: > So let's use `dcterms:source` for that?

I used to do that but note this guidance from [Dublin Core `source`](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/source): "This property is intended to be used with non-literal values. The described resource may be derived from the related resource in whole or in part. Best practice is to identify the related resource by means of a URI or a string conforming to a formal identification system.".

So lots of people use `dcterms:source` not as intended, it's an ObjectProperty, not an AnnotationProperty.

So what to use for annotations? `dcterms;provenance`? Happy to consider options other than `dcterms:source`.

1.1/terms_and_defns.ttl
nicholascar: Some tooling (e.g. TopBraid stuff) likes it, but it doesn't have to be. I can strip this out.

1.1/terms_and_defns.ttl
```diff
@@ -0,0 +1,159 @@
...
+@prefix td: <http://www.opengis.net/def/geosparql/td/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+cs:
+    a owl:Ontology , skos:ConceptScheme ;
+    dcterms:created "2021-06-22"^^xsd:date ;
+	dcterms:creator "OGC GeoSPARQL Standards Working Group" ;
+    dcterms:modified "2021-06-22"^^xsd:date ;
+    dcterms:provenance "Each term (Concept) listed here indicates its own provenance since they are derived from other vocabularies and non-vocabulary sources."@en ;
```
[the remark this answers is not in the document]
nicholascar: True, it's some other kind of note. I'll change.

1.1/terms_and_defns.ttl
nicholascar: Please just put them in!

1.1/terms_and_defns.ttl
nicholascar: We can use a Collection as an unordered bag (there is OrderedCollection too). I don't think we want hierarchy: of this whol vocab, there will be "domains" of terms like perhaps _Semantic Web_ & _Spatial_ etc. but the terms in them are unlikely to be hierarchically defined. Currently some of the terms here clearly have no hierarchical relationship, e.g.  `td:Subject`, `td:Class`, but are nevertheless still in the _Semantic Web_ domain.

I will add in hierarchy where appropriate (perhaps `td:Property skos:narrower td:ObjectProperty` etc. but hierarchies like that just won't cater for all the domain relations.

1.1/terms_and_defns.ttl
nicholascar: No, see above. Can add in hierarchy where appropriate. Just haven't got too many obvious ones for now (other than the suggestion above)

nicholascar:
I don't agree with  modelling that says something like `td:ObjectProperty` is a narrower term than `SemanticWeb` or perhaps `Polygon` is a narrower term than `GIS`. `SemanticWeb` & `GIS` are domains (of science/work etc.) where as `td:ObjectProperty` & `Polygon` are specific things within the domains.

Better is to apply hierarchy when obvious (as above: perhaps `Property` & `ObjectProperty` but then to group related terms in a non-hierarchical way: `Collections`.

1.1/terms_and_defns.ttl
nicholascar: removed in Commit https://github.com/opengeospatial/ogc-geosparql/pull/165/commits/ecd8b78b1877170f1a34e3fa407aa980cd0978ba

1.1/terms_and_defns.ttl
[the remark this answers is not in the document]
nicholascar: Changed to `skos:note` in Commit https://github.com/opengeospatial/ogc-geosparql/pull/165/commits/88e155d00fa3aae53bd07c45d09683a75bdf8454

1.1/terms_and_defns.ttl
nicholascar: I've added hierarchy for at least the Resource & Property terms.

nicholascar: See Commit https://github.com/opengeospatial/ogc-geosparql/pull/165/commits/09dbec599f1abc464554c98d7b2ddbc68054bef3

nicholascar:
> ...probably best to use a generic skos:note.

I've made that change in Commit https://github.com/opengeospatial/ogc-geosparql/pull/165/commits/1e8b83da52be4fefb528865e8dc872d76cb8de35.

> The ISO/IEC directives has guidance about what constitutes a good definition...

I've removed full stops and definite articles in the definitions i Commit https://github.com/opengeospatial/ogc-geosparql/pull/165/commits/1e8b83da52be4fefb528865e8dc872d76cb8de35. They were just copied directly from RDF, RDFS etc...

1.1/terms_and_defns.ttl
nicholascar: Added hierarchy for some of the terms

1.1/terms_and_defns.ttl
FransKnibbe: If I look at the examples, it seems that the annotation literals can be replaced by IRIs. For example, in 
`td:Property dcterms:provenance "From the RDF Schema vocabulary (RDFS)"@en `, couldn't the IRI of the RDF Schema vocabulary be used? I think we should be safe to assume that all definition sources are available online somehow, with a persistent IRI.

1.1/terms_and_defns.ttl
FransKnibbe: Am I right that you are saying that we want to capture different kinds of generalisation relationships? i.e.

1. terms that are defined within a knowledge domain (e.g. Feature is defined in the geoinformatics domain)
2. terms that are more specialised forms of other terms (e.g. Coordinate Reference Systems are a subset of Spatial Reference Systems)

I think it is possible to view both in the same way, as relationship between subsets and supersets. And that means only one mechanism for specifying generalisation relationships needs to be used. A compact example:
```
:GeosparqlTerms 
  a skos:ConceptScheme;
  skos:definition "Terms (and their definitions) used in the GeoSPARQL specification.";
  skos:hasTopConcept :SemWebTerms, :GeoinformaticsTerms
.
:SemWebTerms
  a skos:Concept;
  skos:definition "Terms (and their definitions) from the Semantic Web domain." ;
  skos:inScheme :GeoSparqlTerms
.
:Property
  a skos:Concept;
  skos:broader :SemWebTerms;
  skos:narrower :ObjectProperty, :DataProperty;
  skos:inScheme :GeoSparqlTerms
.
:SRS
  a skos:Concept;
  skos:broader :GeoinformaticsTerms;
  skos:narrower :CRS;
  skos:inScheme :GeoSparqlTerms
.
```
In all cases, `skos:broader` means "is a subset of" and skos:narrower means "is a superset of".

An alternative could be to model the set of terms from the Semantic Web and geoinformatics as concept schemes too, like below:

```
:GeoSparqlTerms 
  a skos:ConceptScheme;
  skos:definition "Terms (and their definitions) used in the GeoSPARQL specification.";
.
:SemWebTerms
  a skos:ConceptScheme;
  skos:definition "Terms (and their definitions) from the Semantic Web domain." ;
.
:SRS
  a skos:Concept;
  skos:narrower :CRS;
  skos:inScheme :GeoSparqlTerms, :SemWebTerms;
.
:Property
  a skos:Concept;
  skos:narrower :ObjectProperty, :DataProperty;
  skos:inScheme :GeoSparqlTerms, :SemWebTerms;
.
```

I think both ways of modelling are expressive enough without `skos:Collection`s. The second approach also has room for terms that are used in GeoSPARQL but do not really belong to either the semweb or geoinformatics knowledge domain.

nicholascar:
> Then how about the second proposal, to have three concept schemes to denote groups of terms (and use skos:inScheme to state to which group(s) a term belongs)?

I don't think we should do this: our goal in this T&D vocab is to list all of the terms that are used in the Standard that users might be unfamiliar with. Since this standard cuts across spatial & semantic, we can expect some semantic terms to be unfamiliar to spatial people & _vice versa_. We don't really need to spend time sorting hierarchies of terms, and certainly not multiple Concepts Schemes of them, we really just need to list them. Multiple CSes make for harder management than one CS matching one governance regime, which is what we have here: all the terms that are needed for GeoSPARQL 1.1, however they are related conceptually. This is far more a flat glossary than it is a taxonomy. Yes we can include some hierarchy where it's obvious - `Property` &gt; `Object Property` etc. - but let's not focus on that at all and lets focus on making just a list of the "top 25 likely to be misunderstood terms" from spatial word, semantic work and any other. Remember, almost all of the terms we will likely list here have home definitions in ontologies/vocabularies that define them, so we don't want to reinvent all those wheels.

Some suggestions for other "top 25 likely to be misunderstood terms" from me:

Term | Note
--- | ---
Profile | we use a particular definition of _profile_ for the characterization of GoeSPARQL 1.1. as a Profiles Vocabulary _profile_. Since there are many uses of the word "profile", we should describe what we are doing.
Feature | We have an ontology definition and yet the term is generic and may be used in text informally
Polygon, Point etc | We should just refer to the definitions we use (SF vocab) but including them in the main T&D probably helps users
Collection | Obviously, given our collection/container discussions
Triplestore | And any other technical implementation we may refer to that perhaps not all users will know. I will check to see if we have referred to this term

Any others you can think of off the top of your head?

FransKnibbe:
> I don't think we should do this: our goal in this T&D vocab is to list all of the terms that are used in the Standard that users might be unfamiliar with. Since this standard cuts across spatial & semantic, we can expect some semantic terms to be unfamiliar to spatial people & _vice versa_. We don't really need to spend time sorting hierarchies of terms, and certainly not multiple Concepts Schemes of them, we really just need to list them. Multiple CSes make for harder management than one CS matching one governance regime, which is what we have here: all the terms that are needed for GeoSPARQL 1.1, however they are related conceptually.

It seems we agree that it is useful to be able to differentiate between the two knowledge domains (semweb and GIS), and to indicate superset/subset relationship relationship in some cases (CRS is a kind of SRS, geodetic datum is a kind of datum, ...). I think especially for the GIS terms it will be useful to express such hierarchy because in practice those terms are used loosely.

As for how to best represent the two knowledge domains. I thought a case could be made to cast them as concept schemes. It seems more in line with the intent of SKOS,  especially when considering that a SKOS concept scheme is an open thing which does not need to be complete (we don't need to define all semweb or GIS terms).  We could use collections instead of concept schemes, but I do not see how concept schemes would require more management. But I admit that I can not think of a good reason not to use collections. So let's stick with that.

The main purpose of this PR is to define the basic setup/template for the terms & definitions vocabulary. Shall I wait for it to be merged into master before making additions?

FransKnibbe:
> Feature       We have an ontology definition and yet the term is generic and may be used in text informally

I think we should use the terms & defs vocab only for terms that are not defined in the ontology (and clearly state that in the terms & defs vocab). In the documentation, we should be very careful to use the correct capitalisation for ontology terms (e.g. "An interesting feature of GeoSPARQL is that it can be used to share data about Features").

To make things even clearer: now that we will have an online vocabulary for terms and definitions we can use term IRIs for hyperlinks to definitions. Not only does it allow direct lookups of definitions, also just the fact that e.g. the word "profile" or "collection" is formatted as a hyperlink shows the reader that the word has a specific meaning.

And, of course, terms that are defined in the ontology could be hyperlinked too from the documentation.

dr-shorthair:
There has been an expectation that OGC standards have a family resemblance to ISO standards, where a Terms and Definitions clause is required. The is a preference for not repeatign definitions, but that then begs the question about where defs are normatively presented, and how they are formalized. I certainly agree that definition text must be synced, but I do think we need all defs in the document somehow, even if those are scraped from the RDF.

jabhay:
It is proposed that we keep this simple (e.g., limited hierarchy, main classes, key concepts, etc.,) in the first instance. Any more complicated modelling should be deferred to a future release.

rob-metalinkage:
I concur with keeping it simple - but remember the goal is simplicity for the end user not the publisher.

Consequently, its vital that you dont proliferate terms - and make the user work hard to work out if this is the same as something similar.  All you need to do is to retain the authoritative URI for the term in the section and use that to retrieve definitions during document construction...  

This pattern will work fine in the Definitions Server - though we have a little work to do to allow terms to show they participate in multiple concept schemes (which is legal).

You can also use SKOS Collections to group them by domain (with overlaps if you really wish) - again we'll do more to show collections heirarchies automatically in the Defs server UI - but your concern is just the doc generation which may group using these.

i.e. dont copy/paste definitions without the normative URI ...

PR #166, opened 2021-06-22
Provide namespace prefixes for all /req/ & /conf/ IRIs

nicholascar:
This PR updates the Clause 5 schedule of prefixes & namespaces and introduces them for Requirements Classes, `georeq`, & Conformance Classes `geoconf`.

Until now, the IRIs for Requirements Classes & Conformance Classes  had no scheme (e.g. `http`) or authority (e.g. `opengis.net`).

### DIFF
--- a/1.1/spec/08-Part-05.adoc
+++ b/1.1/spec/08-Part-05.adoc
@@ -30,19 +30,28 @@ The following IRI namespace prefixes are used throughout this document:
 
 [frame=none, grid=none, cols="1, 6"]
 |===
-| ogc: | http://www.opengis.net/
 | eg: | http://example.com/
-| geo: | http://www.opengis.net/ont/geosparql#
-| geof: | http://www.opengis.net/def/function/geosparql/
-| geor: | http://www.opengis.net/def/rule/geosparql/
-| sf: | http://www.opengis.net/ont/sf#
-| skos: | http://www.w3.org/2004/02/skos/core#
 | gml: | http://www.opengis.net/ont/gml#
 | my: | http://example.org/ApplicationSchema#
-| xsd: | http://www.w3.org/2001/XMLSchema#
+| ogc: | http://www.opengis.net/
+| owl: | http://www.w3.org/2002/07/owl#
 | rdf: | http://www.w3.org/1999/02/22-rdf-syntax-ns# 
 | rdfs: | http://www.w3.org/2000/01/rdf-schema#
-| owl: | http://www.w3.org/2002/07/owl#
+| skos: | http://www.w3.org/2004/02/skos/core#
+| xsd: | http://www.w3.org/2001/XMLSchema#
+|===
+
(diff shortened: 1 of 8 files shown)


situx:
I am not sure if it can work this way.
Section 2 Conformance of the GeoSPARQL 1.0 standard states:
"Requirements and conformance test URIs defined in this document are relative to
http://www.opengis.net/spec/geosparql/1.0"

That would force requirement and conformance URIs to follow this schema also for GeoSPARQL 1.1 right?
e.g. http://www.opengis.net/spec/geosparql/1.0/req/core/sparql-protocol
for the first requirement.

nicholascar:
> Section 2 Conformance of the GeoSPARQL 1.0 standard states:
> "Requirements and conformance test URIs defined in this document are relative to
> http://www.opengis.net/spec/geosparql/1.0"

Oops, didn't see that! A pity they are version locked with /1.0/ in there...

> if we change to a different prefix we should provide the spec version somewhere in the URI.

Perhaps easiest to use the /1.0/ for most existing requirements and just a /1.1/ variant for all new ones. I suppose then, as per design, it will be easy to tell which version of the Spec/Ontology the requirement comes from and since 1.1 inherits all 1.0 Requirements, this will be fine.

> I have created a sample file on how I think requirements should be modeled

```turtle
<http://www.opengis.net/ont/geosparql#Requirement> rdf:type owl:Class .
```

I don;'t think we should have this - introducing a new class to GeoSPARQL for a generic OGC thing. The Mod Spec guides Requirements & Tests, so we should rely on a non-GeoSPARQL definition for this type of thing.

(There isn't such a class definition in OWL perhaps, so let's just use `skos:Concept` and ask the NA to make us a definition for this)

> I would really like to connect requirements between versions in RDF, so that if we conduct a compliance test for GeoSPARQL 1.0 we can infer which requirements of GeoSPARQL 1.1 are fulfilled now as well.

Why don't you set up your RDF file here as yet another SKOS vocab in 1.1, using just `skos:Concepts` and not `geo:Requirements` for now and create two Collections in it: for 1.0 and 1.1? The 1.0 Collection can contain only 1.0 terms, the 1.1 Collection can contain all 1.0 and 1.1 terms?

If you do that, which will also sort out the URIs, I'll kill this PR.

(might even be able to auto-generate Annex A from the vocab...)

situx:
I can do this, but I would also introduce relations between the reqiurements. For example many requirements have been altered and this state should be reflected. E.g. a geosparql 1.0 requirement which is exactly the same to a geosparql 1.1 requirement could get a owl:sameAs relation, an altered requirement could get another relation between them. In this way one couls easily identify how many geosparql 1.1 requirements can be tested in the same way as geosparql 1.0 requirements

situx:
@nicholascar Have a look at #168 
I have not yet created collections, but began modelling requirements and conformance classes

nicholascar:
Closing as superseded by PR #168

PR #168, opened 2021-06-23
Modelled requirements

situx:
I have begun to model requirements and conformance classes in RDF.
Still todo:
* Add requirements to GeoSPARQL extensions
* Work out changes in requirements for GeoSPARQL 1.1
* Change rdfs:isDefinedBy linkes in geo.ttl to point to requirements URIs rather to the whole standard

I have created certain properties that are not standard currently to model requirements and conformance classes. I would like to get some feedback if we could replace them with a standardized vocabulary or whether we need to ask the naming authority for a definition.


nicholascar:
> Change rdfs:isDefinedBy linkes in geo.ttl to point to requirements URIs rather to the whole standard

Awesome!

nicholascar:
A few requests:

1. You seem to have accidentally re-added `inSRS` to the ont, so that will have to go again.

2. Can you move from `rdfs:label` to `skos:prefLabel` in `geosparqlreq.ttl`? Just to be consistent with the use of SKOS elsewhere in there.

3. Can you add in a `ConceptScheme` + `Ontology` object in `geosparqlreq.ttl`? OGC will need such document-elvel object to package up and deliver these items.

Once you're done, I'll try building Annex A ASCIIDOC from this.

situx:
Thanks for the suggestions. Would it also be possible to generate a changelog from this data?

nicholascar:
I suppose so, at least to compare to the “new things in 1.1” sections we’ve made manually

situx:
I think the pull request is ready. However, I wonder if we should also modify the geo.ttl file of GeoSPARQL 1.0 to point to the exact GeoSPARQL 1.0 requirements?
Is that allowed or does it change the standard?

nicholascar:
> I wonder if we should also modify the geo.ttl file of GeoSPARQL 1.0 to point to the exact GeoSPARQL 1.0 requirements?
Is that allowed or does it change the standard?

That's a good question. I think we will be bound to not change it, just as we have not changed the 1.0 Spec doc at all. All we've been able to do is to add elements to the 1.0 profile, including the profile definitions itself. 

Consider the 1.1 ontology the updated 1.0 ontology, including all the annotations that would be backported to the 1.0 version if allowed, i.e. not just additions but enhancements such as what you suggest.

situx:
I have built the vocabulary documentation here: https://situx.github.io/ogc-geosparql/ in case it is better to review.

nicholascar (approved):
Nothing wrong with this PR but there is something wrong with pyLODE: it's not rendering the labels of the Collection objects in the vocab. Will have to be fixed

situx:
Well, there is something to be added now I guess because collections have now been merged. I will add the requirements now

dr-shorthair (approved):

PR #171, opened 2021-06-30
Update geo.ttl to add a base uri statement

### LINKED ISSUE #147: Base IRI declaration
This is a request for a base IRI declaration in all OWL/RDF files. In turtle, this would appear as an @base declaration at the top of the files.

See also: https://www.w3.org/TR/turtle/#relative-iri

paulc-dstl:
Added @base statement to provide baseuri. Without trailing hash to align with the ontology definition statement. Closes #147

### DIFF
--- a/1.1/funcsrules.ttl
+++ b/1.1/funcsrules.ttl
@@ -1,3 +1,5 @@
+BASE <http://www.opengis.net/def/geosparql/funcsrules> # baseuri added as per issue 147 
+
 PREFIX dcterms: <http://purl.org/dc/terms/>
 PREFIX funcs: <http://www.opengis.net/def/function/geosparql/>
 PREFIX owl: <http://www.w3.org/2002/07/owl#>
--- a/1.1/funcsrules.ttl
+++ b/1.1/funcsrules.ttl
@@ -1,15 +1,15 @@
-BASE <http://www.opengis.net/def/geosparql/funcsrules> # baseuri added as per issue 147 
+@base <http://www.opengis.net/def/geosparql/funcsrules> . # baseuri added as per issue 147 
 
-PREFIX dcterms: <http://purl.org/dc/terms/>
-PREFIX funcs: <http://www.opengis.net/def/function/geosparql/>
-PREFIX owl: <http://www.w3.org/2002/07/owl#>
-PREFIX policy: <http://www.opengis.net/def/metamodel/ogc-na/>
-PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
-PREFIX rules: <http://www.opengis.net/def/rule/geosparql/>
-PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
-PREFIX status: <http://www.opengis.net/def/status/>
-PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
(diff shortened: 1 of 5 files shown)


paulc-dstl:
The other files will also need a base uri adding...

paulc-dstl:
All turtle files now have a base uri added as per commit above. Now really closes #147

nicholascar:
@paulc-dstl what advantage does this addition give us? I'm not disputing that it's possible and legal, RDF/Turtle wise, but what's the benefit?

paulc-dstl:
> @paulc-dstl what advantage does this addition give us? I'm not disputing that it's possible and legal, RDF/Turtle wise, but what's the benefit?

The only benefit I can see is that in some cases where a consumer may wish to use the base uri. If it is not there, it can't be used. And I don't think having it there causes any issues. Although of course, any consumer would need to be aware of the hash syntax of the uris being used across the ontology and accomodate that (it not being a legal part of any base uri), but that is always the case. 

Perhaps @neilotte could add some detail regarding his intended use please...?

paulc-dstl:
@jabhay I think this needs a 2md reviewer; who would be up for that do you think?
@nicholascar do you think the conversation on #147 is sufficient clarification? 

Thanks :-)

1.1/funcsrules.ttl
```diff
@@ -1,3 +1,5 @@
+BASE <http://www.opengis.net/def/geosparql/funcsrules> # baseuri added as per issue 147
```
dr-shorthair, line 1: This is SPARQL syntax. Is this intentional?

paulc-dstl: SPARQL syntax was used for the baseuri as SPARQL syntax had been used for the existing prefix statements. Agreed this is different from all the other ttl files, but either syntax (`PREFIX` and `BASE` vs `@prefix` and `@base`) are valid. Ought we adopt a consistent form across all the ttl files?

dr-shorthair: Well it took me a couple of minutes to figure out why it was different, and I'm an RDF regular! 
I think it would be better if we standardized to the Turtle syntax in Turtle files (even if SPARQL works as well)

paulc-dstl: Agreed. I've amended all the prefix statements and the base statement in funcsrules.ttl to use Turtle syntax. So now all Turtle files in the repo are consistent on this. 
`@prefix prefixName <uri> .` rather than `PREFIX prefixName <uri>` 
and 
`@base <uri> .` rather than `BASE <uri>`

nicholascar: > better if we standardized to the Turtle syntax in Turtle files 

From [the Turtle 1.1 spec](https://www.w3.org/TR/turtle/#h_note_1): "The case-insensitive 'PREFIX' and 'BASE' forms were added to align Turtle's syntax with that of SPARQL.".

I prefer this syntax as it's far easier to copy and paste Turtle to SPARQL and back. Particularly apt in GeoSPARQL given the query + RDF modelling nature of the standard.

RDF tools out there seen to handle this syntax except for... TopBraid composer! So I undertand why @dr-shorthair doesn't use it...

Fair enough to change to the (horrible, old, nasty) `@prefx` style for wider use.

nicholascar (approved):
I've removed comments in the RDF files and merged in master. I had to remove the `geosparql:` prefix declaration as this conflicted with existing prefix use (as per GeoSPARQL 1.0)

Issue #82, opened 2021-02-17
Define precedence order of inCRS v. Literal CRS


nicholascar:
Some geometry literal formats contain CRS indicators but now we also have an `inCRS` property. We need to indicate precendence order in the following places:

1. in the spec
2. in the ontology
    * perhaps with 2 examples for inCRS for literals with, and without CRS indicated

mathib:
We might need to split the discussion here for each case (geometry format). The spatial data on the web best practices contain a useful appendix regarding the use of CRS in existing geoformats: https://www.w3.org/TR/sdw-bp/#applicability-formatVbp. If OK, I'll try to assemble and maintain the findings of the supported geometry formats and their CRS particularities here. Decisions taken by the group will be added, in other words the summary below might change over time.

* ## WKT
  - WKT spec: no default CRS; not possible to include CRS in geometry format; 2D and 3D geometry
  - GeoSPARQL 1.0 (and PostGIS): allow the definition of a CRS inside the literal. Default is WGS84 (2D)
  - GeoSPARQL 1.1: backwards compatibility with GeoSPARQL 1.0 required
     - case 1: CRS both inside literal and with `inCRS` => CRS from literal (backwards compatible) **OR** CRS from `inCRS`
     - case 2: CRS only in literal => CRS from literal
     - case 3: CRS only with `inCRS` => default to WGS84 (backwards compatible) **OR** CRS from `inCRS`
     - case 4: no CRS => default to WGS84
  - conclusion: if we want to maintain backwards compatibility with GeoSPARQL 1.0 (target of 1.1 version), the `inCRS` pattern is of little use for WKT unless for creating a queryable annotation (value of `inCRS` should be equal to CRS of literal or default CRS). We can't even indicate to favor the `inCRS` pattern because of the default to WGS84 in GeoSPARQL 1.0.
  - note: For GeoSPARQL 2.0, where backwards compatibility is less of an issue, the pattern with `inCRS` could entirely replace the current way of modeling (i.e. the CRS inside the literal).
* ## GeoJSON and KML
  - GeoJSON and KML current spec: 'fixed' CRS (WGS84) according to spec but the KML format can in theory be extended for other CRS; 2D (GeoJSON) and 3D (KML)
  - GeoSPARQL 1.0: did not support GeoJSON and KML so no issues with backwards compatibility
  - GeoSPARQL 1.1: 
    - case 1: no CRS => default to WGS84
[...]

PR #173, opened 2021-07-03
adding a bunch of shapes

mathib:
this pull request from my fork adds a series of SHACL shapes in addition to the first batch from https://github.com/opengeospatial/ogc-geosparql/pull/152

Some notes and open questions remain:
- You'll see that I numbered the shapes using TTL comments, to make it easier to reference a shape in the upcoming discussions. We should perhaps also do this at some point also in the URIs to make testing/debugging easier for us, as well as users of the shapes.
- Shape 16 and higher are currently only described briefly in text. Depending on the opinions of the group, one or more of these can be considered. 
- I explicitly named property shapes that belong together with a node shape. The `sh:message` is in those cases placed on the subshape (i.e. the property shape) to allow nicely formatted validation reports. Otherwise, the message is not emitted and replaced by a default piece of text from the validation engine. By explicitly naming the property shapes in these cases, the triggering shape is also clearly given in the validation report
- Shape 1 now also (at least temporarily) has a `sh:maxCount` of 1 following the discussion between @nicholascar , @FransKnibbe (see third discussion on [validator.ttl file in the original pull request](https://github.com/opengeospatial/ogc-geosparql/pull/152/files/9dda38cfbd2fb5daa13c741bc14b936cb545a61a#diff-16ad330b0d5239c0af42084baac2e54af94c8088d47214de52d5ff77bf3eb4a4)). This implies that each `geo:Geometry` node can only have one outgoing `geo:hasSerialization` relation.
- if Shape 1 remains as described above, we might need to consider to remove `geo:hasSerialization` from the list of mutually excluding properties, since a reasoner will always infer the `geo:hasSerialization` relation from the more specific subproperties that are asserted. I believe `geo:hasSerialization` will not be asserted in a dataset? See also the text proposal for shape 31 ("geo:hasSerialization should never be asserted since it underspecifies the geometry serialization")
- in case of Shape 9 (`geo:dggsLiteral` or `geo:auspixDggsLiteral`), I believe there have been some recent changes in the ontology? If the datatype `geo:dggsLiteral` will always be asserted in a dataset, then there should be no more check on the validity of the more specific type of DGGS since anyone can define a new one and it's not anymore modelled as an RDF datatype accompanying a literal?
- If additional geometry metadata properties are introduced, new shapes similar to Shapes 10-14 might be needed 
- we need to indicate if reasoning is to be used or not, before the validation starts. This can be formally done on the level of the shapes graph (i.e. it counts for all shape constraints in the file) and/or in a human-readable manner in the text description of each shape (`rdfs:comment` or `skos:note`). It's also possible to add it to the `sh:message`
- in case we say that the shapes should work without reasoning, we might consider to add shapes to indicate that users asserted redundant triples. E.g. many classes, see Shape 32-35 proposal, can be inferred from the `rdfs:domain` / `rdfs:range` of properties or from `rdfs:subClassOf` of more specific classes.

Note that what concerns me, all shapes can be discussed. I'm not trying to enforce my view on the group.


situx:
"Each node with one or more outgoing relations that are either geo:hasSerialization or a specialisation of it should have exactly one incoming geo:hasGeometry relation."
I have been wondering about this.
Maybe just for my clarification let's assume the following situation:
* We define a school building as ex:school1 rdf:type ex:School . ex:School rdfs:subClassOf geo:Feature .
* We want to assign a geometry to the school instance ex:school1, so we create: `ex:school1 geo:hasGeometry ex:school1_geom . ex:school1_geom geo:asWKT "POLYGON(...)"^^geo:wktLiteral . `
* However: The school is also used as a floodshelter in times of emergencies.
Then what we should do is to create:

ex:school1 rdf:type ex:FloodShelter . ex:FloodShelter rdfs:subClassOf geo:Feature .
`

instead of
ex:floodshelter1 rdf:type ex:FloodShelter . ex:FloodShelter rdfs:subClassOf geo:Feature .
with:
ex:floodshelter1 geo:hasGeometry  ex:school1_geom .

mathib:
`Is that correct?`
Yes, but another pattern is also possible, but with duplication of the literal if it's the same for both the school and the floodshelter (but I guess most triplestores will take care of the double literal while doing indexing):
```
ex:floodshelter1 a ex:FloodShelter ; geo:hasGeometry ex:floodshelter1_geom .
ex:floodshelter1_geom geo:asWKT "POLYGON(...)"^^geo:wktLiteral  .`
```

If the shape you mention is retained, the following however, would return a violation:
```
ex:school1 geo:hasGeometry ex:school1_geom .
ex:school1_geom geo:asWKT "POLYGON(...)"^^geo:wktLiteral .
ex:floodshelter1 geo:hasGeometry ex:school1_geom .
```

situx:
Yes, exactly, I wondered if in those cases we would really need to define a separate geometry literal.
That sounds to me like a redundancy that can be avoided.

Another case could be:
1. Floodshelters without geometry are given in one graph
2. Schools with geometry are given in another graph
3. I want to merge them but some restriction on either school or flood shelter prevents me from using a property on a joined instance of flood shelter and school

Do I need to create redundancies in those cases and is that what we want with GeoSPARQL?

mathib:
if the geometry node and literal is the same, you can also just say that your school1 is both of type `ex:School` and type `ex:Floodshelter`? It all depends if the URI of the school instance node is shared across datasets or not. I guess it is, since you expect the geometry instance node to be shared and reused across datasets?

dr-shorthair:
@mathib this is exactly why Geometry and Feature are separate classes. In the OGC/ISO TC 211 metamodel, an entity ('feature') has a geometry, but a geometry does not have a semantic type other than its geometric type. So an entity may have more than one geometry, and a geometry may be associated with more than one entity, but they are distinct concepts.

rwarren2:
Three additional use cases for the Geometry and Feature separations:

* Features that change locations / geometry over time, such as front lines or country boundaries during a war .
* The definition of Features that are not materialized can be made through the addition / reduction of different Geometries using the within/disjoint/overlap properties. I have a (buggy) set of a few historical examples [here](https://blog.muninn-project.org/node/75) for country boundaries.
* Features can have disputed geometries which can be linked with a prov frameworks to assign point of vue.

situx:
I think in light of these other examples, my recommendation would be to drop this SHACL rule, or at least to discuss it in the next call, as there seem to be a lot of differing opinions about it

mathib:
Regarding the additional comments from @dr-shorthair and @rwarren2 : I fully agree that `geo:Geometry` and `geo:Feature` are different things and the discussed shape does not deny that (nor do the other shapes); in fact, I have created shape 3 `<hasGeometry-hasGeometry>` exactly for this reason. The same for allowing to have multiple `geo:Geometry` nodes connected to a single `geo:Feature` instance: there are multiple use cases for that situation as you point out and I did not define a shape that requires otherwise. Other use cases are derived geometries (incl. separate geometry serializations, but there I'm in ["conflict" with Frans](https://github.com/opengeospatial/ogc-geosparql/pull/152#discussion_r667821000) ;-) see shape 1 `<hasGeometry-hasSerialization>` )

> and a geometry may be associated with more than one entity

This is the central point of the discussion around this shape (shape 2 `<hasSerialization-hasGeometry>`) as discussed with @situx above. It's still unclear why it's something we so desperately need as there are clear alternatives to model the situation.

> at least to discuss it in the next call

Sure, that was actually the goal: to have us all discuss and define what we consider to be the expected modelling patterns with GeoSPARQL. I just provided a starting point for discussion. What about [the other proposed shapes (and placeholder proposals for shapes)](https://github.com/opengeospatial/ogc-geosparql/blob/916f51630f106c1829abd1e2d7db7b735ec85091/1.1/validator.ttl)?

BTW, I just found out that one of the issues around CRS/SRS ([the need to enforce exactly one CRS/SRS to be entered in one move together with the RDF literal containing the geometry](https://github.com/opengeospatial/ogc-geosparql/issues/82#issuecomment-794606687)) can be solved with an appropriate (mandatory) shape. When using SHACL in "guard-mode", such a shape would prevent users from entering a geometry node with a literal that has no CRS/SRS connected. What do you think? In the case of GeoJSON, an additional shape can require that the CRS should be WGS84.

situx:
We have established that we want to only include shapes that check the structural nature of the graph in the last meeting into the standard.
However, we also discussed that there are SHACL constraints that should not be part of the standard, but could be useful to point to, such as GeoSHACL (https://github.com/chrdebru/chrdebru.github.io/tree/master/ontologies/geoshacl)
I have created a new repository to collect those and also possible SHACL inferencing rules.
https://github.com/situx/geoshaclrules
For now, they mainly include the addition of e.g. the geo:isEmpty property to a geometry with the help of the geof:isEmpty function, but maybe you have further ideas on what could be inferenced.
When creating that I stumbled upon a few questions:

- I wanted to create a validation to check if the area assigned to a geospatial feature matches the area of its geometries. So my question here is: Which area does it have to match if we have many geometries assigned to a feature instance? Every geometry that has an area? I also see that we have no restriction on how often geo:hasArea can be assigned
- The same goes for the geo:hasLength property
- Validating the check of the geo:hasBoundingBox property should be ok, because one can check that all geometry instances are included in the possibly many bounding boxes which might be defined

Do you have any guidance here?

nicholascar:
> Validating the check of the geo:hasBoundingBox property should be ok, because one can check that all geometry instances are included in the possibly many bounding boxes which might be defined

I think this sort of check might be getting too constraining and perhaps we don't want to go there and just leave people to implement their own rule for this sort of thing. 

I think GeoSPARQL shouldn't exhaustively apply every spatial validation check, only ontological formulation error checks.

It may be that certain bounding boxes, perhaps high-resolution ones, don't include all the points of low-resolution boundary polygons - who knows: there could be many odd such data that's actually still OK. 

So let's just make shapes that ensure people are formulating GeoSPARQL RDF correctly but let's not look into the spatial data itself. In 1.0, GeoSPARQL didn't much delve into the spatial data.

mathib:
I just uploaded the newest version of validator.ttl containing the shapes, incorporating the results of the discussions during the last two meetings. Below is a full status update of the changes:

* shape 1: under discussion in https://github.com/opengeospatial/ogc-geosparql/issues/177 (similar pattern also in shape 32)
* annotation of SHACL file
	- SHACL constraints (not rules > inferencing)
	- updated base URI and ontology URI (validation instead of validator)
	- no `rdfs:isDefinedBy` in the shapesGraph => users can select whatever shapes they think are useful
	- the shapesGraph node remains an `owl:Ontology` (see range of `sh:suggestedShapesGraph` in https://www.w3.org/ns/shacl.ttl being applied between the SHACL terminology ontology and the SHACL-SHACL shapesGraph)
	- all URIs of shapes are now numbered (renumbering will happen when all shapes are completed and accepted)
* shape 2: the `sh:maxCount` has been removed as decided in the meeting of July 14. This also means that shape 15 had to be removed.
* shape 9: entire shape has been removed as decided in the meeting of July 28 as the `geo:asDGGS` property can go to literals with different datatypes
* shape 15: removed as it forms the counterpart of shape 2
* shape 16-19: 
	- someone (@nicholascar ?) needs to provide regex for each of the literal datatypes (except DGGS). The skeleton of the shapes is provided. I tested with some minimal regex, but I was not able to create the correct validation report with pyshacl (might be an issue with Turtle and/or the Turtle parser of RDFLib? See: https://github.com/RDFLib/pySHACL/issues/47)
	- it seems like it's not possible in SHACL to select RDF literals based on the datatype alone, only via the incoming property which can be generic (`geo:hasSerialization`). RDF literals usign GeoSPARQL datatypes and which are connected using `geo:hasSerialization` or another property which is not a subproperty of `geo:hasSerialization` will not be detected.
	- shape 20 (DGGS literals) is left out as decided in the meeting of July 14, as it appears that the form of DGGS literals is rather undefined (+ we can't select the RDF literal via the datatype)
* shape 21: added
* shape 23-24: separate new shapes are not needed for `geo:hasDefaultGeometry` and `geo:defaultGeometry`
	- extended property shapes of shape 2, 21 and 32
	- added sibling node shapes for shape 1 and 3 (resp shapes 1-b, 1-c, 3-b, 3-c) => the property shapes are reused for each cluster of node shapes
* shapes 25-29 are left out since it seems that SHACL does not allow to access an RDF literal via its datatype
* shape 30: added
* shape 31 is left out as decided in the meeting of July 14
* shape 32: added. Note that it's influenced by the decision for shape 1	
* no shape constraints (e.g. max cardinality) implemented for `geo:hasLength`, `geo:hasArea`, `geo:hasBoundingBox`, etc

situx:
@mathib I think we need the following additional SHACL constraints:
* Checking that a geo:FeatureCollection is only associated with instances which are subClassOf geo:Feature
* Checking that a geo:GeometryCollection is only associated with instances which are subClassOf geo:Geometry
* Checking that a geo:SpatialCollection is only associated with instances which are subClassOf geo:SpatialObject

@nicholascar has used the relation dcterms:isPartOf relation between a FeatureCollection and a Feature in the examples. I am currently not clear about if that is our official recommendation?
Is that so? Then it could be used for the SHACL constraints

situx:
Maybe we could write this in a NOTE next to the requirements of FeatureCollection and/or SpatialObjectCollection that the recommended way to connect a Feature to a FeatureCollection is to use rdfs:member. It is good that it is included in the ontology, but I think it is not immediately obvious from reading the spec only

nicholascar:
> I think it is not immediately obvious from reading the spec only

It is actually there in the Spec, see https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_class_geofeaturecollection

nicholascar:
I've made some minor text improvements to the Shapes. 

If we exclude the REGEX shapes for now, based on our discussion last meeting that those shapes that peer into literals should be added to an extended set of shapes, not included here in the main GeoSPARQL Spec, then I'm ready to finalise this, approve and merge.

@mathib do you want to move Shapes 16-17-18-19-20 our or will you?

@jabhay still has to provide a place in Geosemantics DWG for these extended shapes and also fro other extended examples currently in https://github.com/opengeospatial/ogc-geosparql/tree/master/1.1/examples

situx:
> > I think it is not immediately obvious from reading the spec only
> 
> It is actually there in the Spec, see https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_class_geofeaturecollection

Yes, I know the definition of the class gives it away. However, I think a short paragraph explaining the usage of rdfs:member could be helpful for people reading the spec who are not too proficient in reading owl:Restrictions

nicholascar:
> However, I think a short paragraph explaining the usage of rdfs:member could be helpful

Notes added in https://github.com/opengeospatial/ogc-geosparql/pull/188

1.1/validator.ttl
```diff
@@ -408,7 +408,7 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however
 	a sh:PropertyShape ;
 	sh:path geo:isSimple ;
 	sh:maxCount 1 ;
-	sh:message "The geo:Geometry node with an outgoing geo:isSimple property should have maximum one outgoing geo:isSimple property."@en ;
+	sh:message "A geo:Geometry node should have a maximum one outgoing geo:isSimple property."@en ;
```
mathib, line 411: "a maximum of" ? (returns multiple times in the messages)

mathib:
Hi all, I've updated the SHACL shapes file with the following content:
* shape 1: splitted in 5 parts:
    - shape 1a: a `geo:Geometry` node should have min one outgoing `geo:hasSerialization` relation (or specialization) - `sh:Violation` (default)
    - shape 1b: a `geo:Geometry` node should have max one outgoing `geo:hasSerialization` relation (or specialization) - sh:Warning (as agreed during the last call)
    - shapes 1c - 1f: a `geo:Geometry` node should have max one outgoing `geo:asWKT`/`geo:asKML`/`geo:asGML`/`geo:asGeoJSON` relation - `sh:Violation` (default, as agreed during the last call)
* shapes 15a - 15d: added additional shapes to check metadata properties (`geo:hasMetricResolution`/`geo:hasMetricAccuracy`/`geo:hasSpatialAccuracy`/`geo:hasSpatialResolution`) on `geo:Geometry` nodes, to limit them to max 1
* shapes 33 - 35: the shapes for the `geo:GeometryCollection`/`geo:FeatureCollection`/`geo:SpatialObjectCollection` as requested by Timo

I've also took the freedom to add myself as a contributor to GeoSPARQL 1.1, by adding my name and affiliation in `geo.ttl`. Hope this is OK?

@nicholascar : can you fill in the correct regex for the shapes 16-19? You might also want to renumber the shapes via the URIs, but I'll leave that to you so you can maintain the overview and some coherence in style.

situx:
@mathib Thank you for the update.
I still have two questions about the collection shapes:
1. I think it would make sense to declare that a connection between e.g. a Feature and a FeatureCollection could also be made with a sub-property of rdfs:member. Is that covered by the current rules? If not I think it would make sense to change this.
2. I compared your rule to the example created by @nicholascar and found that he defined FeatureCollection rdfs:member Feature relations, whereas you defined Feature rdfs:member FeatureCollection relations in the rule. I looked up the specification of rdfs:member and I think @mathib is correct, so we would need to change the example, I think?

nicholascar:
>  I think it would make sense to declare that a connection between e.g. a Feature and a FeatureCollection could also be made with a sub-property of rdfs:member. Is that covered by the current rules? If not I think it would make sense to change this.

+1

> I compared your rule to the example created by @nicholascar and found that he defined FeatureCollection rdfs:member Feature relations, whereas you defined Feature rdfs:member FeatureCollection relations in the rule. I looked up the specification of rdfs:member and I think @mathib is correct, so we would need to change the example, I think?

I think I'm correct here! If you look at [the definition in the Spec](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_class_geofeaturecollection) that @dr-shorthair created we have:

```
geo:FeatureCollection
...
  rdfs:subClassOf [
      rdf:type owl:Restriction ;
      owl:allValuesFrom :Feature ;
      owl:onProperty rdfs:member ;
    ] ;
...
```

So it's `<FeatureCollection_X> rdfs:member <Feature_Y> .`, not `<Feature_Y> rdfs:member <FeatureCollection_X>`.

Also, looking at the RDFS ontology, the definition for `rdfs:member` reads "A member of the subject resource." so it's the _object_ of the triple with the predicate `rdfs:member` that belongs to the container that is the _subject_ of that triple.

situx:
Okay in that case the SHACL shape needs to be modified

nicholascar:
> Okay in that case the SHACL shape needs to be modified

Done

situx (approved):
I think the pull request is good to go.
Not sure about the regex though, but I would leave it to Nick to decide on that.
The regex could also be a different PR.

nicholascar:
I'm going to merge this in now so that the shapes file appears in the master branch of the repo as we need to show it off to people. But the shapes still need work, so we will need more PRs here

PR #176, opened 2021-07-12
Modeled conformance classes

situx:
Modeling of conformance classes defined in the GeoSPARQL standard.
This pull request waits for consent about how to represent URIs of requirements, probably to be solved in the next GeoSPARQL meeting call


nicholascar:
@situx I've made no content changes to the vocab, just appled prefixes so we can more easily see the IRIs that have been defined. Also, I added the vocab to the Profile.

situx:
Ok, I removed all non-relevant GeoSPARQL 1.1 URIs

situx:
I went through all requirements, found that the geof:transform function was misplaced and corrected that.
Then, I renumbered every requirement to be represented using the ascending order from GeoSPARQL 1.0.
So this is the provisional numbering we could follow. If we decide otherwise that would be up to us

situx:
Updated the spec with the new URIs

1.1/profile.html
```diff
@@ -364,7 +364,7 @@ <h3>Generator script for GeoSPARQL Rules in RIF</h3>
     <table>
         <tr>
           <th><a class="proplink" href="https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/description">Description</a></th>
-          <td>This is a Python script used to generate the 'GeoSPARQL Rules in RIF' artifact from templates specified in the 'Ontology specification in PDF'</td>
+          <td>This is a Python script used to generate the 'GeoSPARQL Rules in RIF' artifact from templates specified in the 'Ontology specification in PDF'. This script is not indendedto be executable 'on the web'. It's content is presented as plain text for reference</td>
         </tr>
```
FransKnibbe, line 368: intendedto > intended to
It´s > Its

1.1/profile.ttl
```diff
@@ -104,40 +104,47 @@
...
 
 <http://www.opengis.net/def/geosparql/rules-script>
     dcterms:title "Generator script for GeoSPARQL Rules in RIF"@en ;
-    dcterms:description "This is a Python script used to generate the 'GeoSPARQL Rules in RIF' artifact from templates specified in the 'Ontology specification in PDF'"@en ;
-    dcterms:conformsTo <https://www.w3.org/TR/rif-bld/> ;
-    dcterms:format "text/plain" ;  # it is not indended that the script be executable "on the web" only its content visible for reference
-    prof:hasArtifact "https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/rifgenerator.py"^^xsd:anyURI ;  # to be updated with GeoSPARQL 1.0 publication
-    prof:hasRole role:guidance ;  # since it is an informative artifact
+    dcterms:description "This is a Python script used to generate the 'GeoSPARQL Rules in RIF' artifact from templates specified in the 'Ontology specification in PDF'. This script is not indendedto be executable 'on the web'. It's content is presented as plain text for reference"@en ;
+    dcterms:format "text/plain" ;
```
FransKnibbe, line 130: intendedto > intended to
It´s > Its

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+                     rdfs:seeAlso reqs10gx:gml-literal ;
+                     skos:definition "Verify that queries involving geo:gmlLiteral values return the correct result for a test dataset" ;
+                     skos:prefLabel "GeoSPARQL 1.0 Conformance Test: GML Literal"@en .
+
+conf10gx:gml-literal-empty rdf:type owl:NamedIndividual ,
+                                    skos:Concept ;
+                           geo:testPurpose "check conformance with this requirement" ;
+                           geo:testType geo:Capabilities ;
+                           rdfs:seeAlso reqs10gx:gml-literal-empty ;
+                           skos:definition "verify that queries involving empty geo:gmlLiteral values return the correct result for a test dataset" ;
```
FransKnibbe, line 200: verify > Verify

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+@prefix reqs10core: <http://www.opengis.net/spec/geosparql/1.0/req/core/> .
+@prefix reqs11core: <http://www.opengis.net/spec/geosparql/1.1/req/core/> .
+
+@base <http://www.opengis.net/def/geosparql/reqs> .
+<http://www.opengis.net/def/geosparql/reqs> rdf:type owl:Ontology ;
+                                             dcterms:created "2021-06-28"^^xsd:date ;
+                                             dcterms:modified "2021-07-13"^^xsd:date ;
+                                             rdfs:seeAlso <http://www.opengis.net/doc/geosparql/1.1> ;
+                                             skos:definition "A vocabulary of the requirements and conformance classes in the GeoSPARQL 1.1 specification"@en ;
+                                             skos:prefLabel "GeoSPARQL 1.1 Requirement and Conformance Class Register" .
```
FransKnibbe, line 45: Requirement > Requirements

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+
+skos:prefLabel rdf:type owl:AnnotationProperty .
+
+prov:wasDerivedFrom rdf:type owl:AnnotationProperty .
+
+skos:Collection rdf:type owl:Class .
+
+skos:Concept rdf:type owl:Class .
+
+skos:ConceptScheme rdf:type owl:Class .
```
FransKnibbe, line 71: I think lines 49-71 can be removed. Those terms are defined elsewhere.

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+
+<http://www.opengis.net/def/geosparql/reqs> rdf:type owl:NamedIndividual ,
+                                                     skos:ConceptScheme .
+
+confs10:core rdf:type owl:NamedIndividual ,
```
FransKnibbe, line 76: What is the reason behind using owl:NamedIndividual here (and at other places)?

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+@prefix conf11core: <http://www.opengis.net/spec/geosparql/1.1/conf/core/> .
+@prefix reqs10core: <http://www.opengis.net/spec/geosparql/1.0/req/core/> .
+@prefix reqs11core: <http://www.opengis.net/spec/geosparql/1.1/req/core/> .
+
+@base <http://www.opengis.net/def/geosparql/reqs> .
+<http://www.opengis.net/def/geosparql/reqs> rdf:type owl:Ontology ;
+                                             dcterms:created "2021-06-28"^^xsd:date ;
+                                             dcterms:modified "2021-07-13"^^xsd:date ;
+                                             rdfs:seeAlso <http://www.opengis.net/doc/geosparql/1.1> ;
+                                             skos:definition "A vocabulary of the requirements and conformance classes in the GeoSPARQL 1.1 specification"@en ;
```
FransKnibbe, line 44: Change to "... requirements classes and conformance classes ..."?
And shouldn't sentences like these have proper punctuation? That is something we need to agree on for all RDF code. I have just extended [the project wiki](https://github.com/opengeospatial/ogc-geosparql/wiki) to be able to set text formatting guidelines.

FransKnibbe:
I don't see modular dependencies represented in reqs.tll. But I do think that is relevant information. Could we use [dcterms:requires](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/requires) for those relationships? For example:
`confs10:topology-vocab-extension dcterms:requires confs10:core`

FransKnibbe:
Capitalisation seems to be inconsistent in `reqs.tll`: _requirements class_ and _conformance class_ are sometimes capitalised and sometimes not. I think it would be best not to capitalise those terms, as is done in the GeoSPARQL specification document.

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+@prefix reqs11qre: <http://www.opengis.net/spec/geosparql/1.1/req/query-rewrite-extension/> .
+@prefix reqs11ree: <http://www.opengis.net/spec/geosparql/1.1/req/rdfs-entailment-extension/> .
+@prefix reqs11tve: <http://www.opengis.net/spec/geosparql/1.1/req/topology-vocab-extension/> .
+@prefix conf10core: <http://www.opengis.net/spec/geosparql/1.0/conf/core/> .
+@prefix conf11core: <http://www.opengis.net/spec/geosparql/1.1/conf/core/> .
+@prefix reqs10core: <http://www.opengis.net/spec/geosparql/1.0/req/core/> .
+@prefix reqs11core: <http://www.opengis.net/spec/geosparql/1.1/req/core/> .
+
+@base <http://www.opengis.net/def/geosparql/reqs> .
+<http://www.opengis.net/def/geosparql/reqs> rdf:type owl:Ontology ;
```
FransKnibbe, line 40: why not also a skos:ConceptScheme?

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+                                          skos:Concept ;
+                                 skos:member conf10gtx:eh-query-functions ,
+                                             conf10gtx:rcc8-query-functions ,
+                                             conf10gtx:relate-query-function ,
+                                             conf10gtx:sf-query-functions ;
+                                 skos:prefLabel "GeoSPARQL 1.0 Conformance Class: Topology Vocabulary Extension"@en .
+
+conf10core:feature-class rdf:type owl:NamedIndividual ,
+                                  skos:Concept ;
+                         geo:testPurpose "check conformance with this requirement" ;
```
FransKnibbe, line 125: (These comments apply to all similar occurences)

1. The phrase "this requirement" is unclear. How about "Check conformance with the requirement referred to by the subject of the rdfs:seeAlso property of this resource", or something similar?
2.` geo:testPurpose` seems undefined. And the GeoSPARQL ontology does not seem the right place to define such things. It would be better to use a general OGC vocabulary for that. That goes for `geo:testType` and similar terms too.
3. If a more general vocabulary for conformance testing would be made, probably a better way than rdfs:seeAlso can be used to link a conformance test to a requirement.

Perhaps we need a new issue to deal with the necessity of a conformance testing vocabulary at the OGC level.

1.1/reqs.ttl
```diff
@@ -0,0 +1,951 @@
...
+                                     geo:testType geo:Capabilities ;
+                                     skos:definition "Verify that queries involving geo:GeometryCollection return the correct result on a test dataset" ;
+                                     skos:prefLabel "GeoSPARQL 1.1 Conformance Test: Geometry Collection Class"@en .
+
+conf11core:sparql-protocol rdf:type owl:NamedIndividual ,
+                                    skos:Concept ;
+                           geo:testPurpose "check conformance with this requirement" ;
+                           geo:testType geo:Capabilities ;
+                           rdfs:seeAlso reqs11core:sparql-protocol ;
+                           skos:definition "verify that the implementation accepts SPARQL queries and returns the correct results in the correct format, according to the SPARQL Query Language for RDF, the SPARQL Protocol for RDF and SPARQL Query Results XML Format W3C specifications" ;
```
FransKnibbe, line 642: verify > Verify

FransKnibbe (changes requested):
A very welcome contribution. I did see some issues here and there.

1.1/reqs.ttl
situx: Yes, you are correct, there are vocabularies that need to be defined. I think the OGC Naming Authority will do this?
And yes I agree that "this requirement" is not very precise.
However, it is exactly the text that is written in Annex A. So either we decide to copy this text 1:1 or we use a different description. For now, I used the exact definitions from Annex A.

1.1/reqs.ttl
situx: I guess it is a matter of discussion. I somehow had in mind that skos definitions were not supposed to end with a dot

1.1/reqs.ttl
situx: This had something to do with pylode being able to generate the documentation properly. It may be obsolete now that is is also classified as ConceptScheme. I would need to check that

situx:
> I don't see modular dependencies represented in reqs.tll. But I do think that is relevant information. Could we use [dcterms:requires](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/requires) for those relationships? For example:
> `confs10:topology-vocab-extension dcterms:requires confs10:core`

Yes, we could add this. The question is: Do we add this on a requirement basis or on an extension basis.
I would prefer on a requirement basis and would look through all the requirements that build upon previously defined requirements

1.1/reqs.ttl
FransKnibbe: > Yes, you are correct, there are vocabularies that need to be defined. I think the OGC Naming Authority will do this?

The Naming Authority does come to mind. Maybe we can cooperate - propose a vocabulary for the NA to publish? @nicholascar will probably have an idea about a good approach. 

> And yes I agree that "this requirement" is not very precise.
> However, it is exactly the text that is written in Annex A. So either we decide to copy this text 1:1 or we use a different description. For now, I used the exact definitions from Annex A.

When viewed as text the relationship is much clearer (¨requirement¨ printed in boldface). It will probably help to have a vocabulary for requirements/conformance classes. It could contain a property to relate a conformance test to a requirement.

FransKnibbe:
> Yes, we could add this. The question is: Do we add this on a requirement basis or on an extension basis.
> I would prefer on a requirement basis and would look through all the requirements that build upon previously defined requirements

Do you mean for reuirements to indicate required requirements? Or required modules? In GeoSPARQL 1.0 dependecies are made between modules. Wouldn´t it get messy (for us and for implementers) to have more refined dependencies?

situx:
Could be. Ok then, for now I will add dependencies between modules

1.1/reqs.ttl
situx: Correct. For now, this is only achieved by the rdfs:seeAlso statement which might be replaced by another vocabulary once we define one

1.1/reqs.ttl
situx: Changed

1.1/reqs.ttl
situx: I removed it for now

1.1/reqs.ttl
situx: Removed

1.1/reqs.ttl
situx: Corrected

1.1/reqs.ttl
situx: Corrected

1.1/profile.ttl
situx: Corrected

1.1/profile.html
situx: Corrected

1.1/reqs.ttl
situx: Changed it to ConceptScheme

1.1/reqs.ttl
nicholascar: I tend not to use training full stop but often have structured literals for which I need to include internal punctuation e.g.:

```turtle
"""This is a sentence.

This is a second.

No training full stop here"""@en
```

nicholascar:
Approving now and happy to assist with the required additional Naming Authority things, as needed.

nicholascar (approved):

1.1/reqs.ttl
FransKnibbe: What is the reason for omitting the full stop?

FransKnibbe (approved):
I approve, but for the sake of continuity it does seem we need a new issue for solving the case of the missing vocabulary for requirements and conformance tests.

situx:
Okay, I created issue #181 for that and will merge the results for now.

Issue #175, opened 2021-07-11
clearly define 'requirements class' and 'conformance class'


FransKnibbe:
The GeoSPARQL specification uses the terms  'requirements class' and 'conformance class' and says they correspond, but clear definitions of those terms are lacking. Explanations could be added in the text when those terms are used for the first time.

See [issue 120](https://github.com/opengeospatial/ogc-geosparql/issues/120) for discussion on the meaning of the terms.

dr-shorthair:
Those terms are defined in the OGC Policy document - 
"The Specification Model — A Standard for Modular specifications" available from https://www.ogc.org/ogc/policies/directives

FransKnibbe:
Thank you @dr-shorthair. 

_The Specification Model — A Standard for Modular specifications_ provides the following definitions of "conformance class" and "requirements class" (other definitions need to be included because they are referred to):

1. **requirements class**: aggregate of all **requirement modules** that must all be satisfied to satisfy a **conformance test class**.
2. **requirements module**: aggregate of requirements and recommendations of a specification against a single **standardization target type**.
3. **standardization target type**: type of entity or set of entities to which the requirements of a standard apply.
4. **conformance test class**: set of **conformance test modules** that must be applied to receive a single **certificate of conformance**.
Notes: 
    - When no ambiguity is possible, the word "test" may be left out, so **conformance test class** maybe called a **conformance class**.
    - In this standard, the set of requirements tested by the conformance tests within a **conformance class** is a **requirements class** and its dependencies. Each optional requirement will be in a separate requirements class with other requirements that are part of the same option. Each requirements class corresponds to a separate conformance class. In other words, the only options in a specification conformant to this standard will be if a particular requirements class is tested. Each requirements class will be in a 1 to 1 correspondence to a similarly named conformance class that tests all of the requirements class’s requirements.
 5. **conformance test**: test, abstract or real, of one or more requirements contained within a standard, or set of standards.
 6. **certificate of conformance**: evidence of conformance to all or part of a standard, awarded for passing one or more of the conformance test classes specified in that standard.
 Notes:
    - "Certificates" do not have to be instantiated documents; having proof of passing the conformance test class is sufficient. For example, the OGC currently keeps an online list of conformant applications at http://www.opengeospatial.org/resource/products.
 7. **conformance test module**: set of related tests, all within a single conformance test class.

I assume that "requirement module" and "requirement**s** module" are synonyms. 

[...]

PR #180, opened 2021-07-31
Patch issue 175

FransKnibbe:
Adds explanations of "requirements class" and "conformance class". Resolves [issue 175](https://github.com/opengeospatial/ogc-geosparql/issues/175).


situx:
In general I think the changes are ok, but I stumbled upon this part while reading:
 Requirements and conformance test IRIs defined in this document are relative to the namespace IRI `+http://www.opengis.net/spec/geosparql/1.1/+`.
Should we not add the 1.0 URI here as well?

FransKnibbe:
> In general I think the changes are ok, but I stumbled upon this part while reading:
> Requirements and conformance test IRIs defined in this document are relative to the namespace IRI `+http://www.opengis.net/spec/geosparql/1.1/+`.
> Should we not add the 1.0 URI here as well?

Well spotted! I have just changed the text to include both base IRIs.

nicholascar (approved):

situx (approved):
Ok, with these changes, I vote approve!

PR #186, opened 2021-08-11
Internal IRIs

### LINKED ISSUE #183: Use Fragment IRIs for all Spec obejct refs in Spec
Instead of linking a thing defined in the Spec and the ontology, e.g. `Feature` to the ontology, link it to the relevant place in the Spec.

This will change links like this:

```
http://www.opengis.net/ont/geosparql#Feature[`geo:Feature`]
```

to links like this:

```
#_class_geofeature[`geo:Feature`]
```

While I understand the desire to link to a point-of-truth, the links to the ontology "don't work" for humans in the sense that they are fragment IRIs pointing to a single text (turtle) document so always prompt for full document download. Relative hyperlinks back to places in the Spec would be better for human use, as shown in W3C specs, like DCAT, e.g.:

See under the Catalog class section (https://www.w3.org/TR/vocab-dcat/#Class:Catalog), there are links to other things in the ontology & spec, e.g. Catalog Record which is given like this:

```
<a href="#Property:catalog_catalog_record">catalog record</a>
```

nicholascar:
This PR converts all IRIs in the Spec for Class, Property, Datatype & many Functions to internal IRIs that refer to their Spec section (their normative definition) except for the first IRI in each of those definitional sections that refer to the ontology element.

This is so all links, except for the first ones in definitional sections, link back to other places in the document. This makes it much easier for a human to navigate around the document.

I've also made many small updates to Turtle formatting and notes for consistency.

The Spec, rendered:

* https://raw.githack.com/opengeospatial/ogc-geosparql/spec-iris/1.1/spec/spec.html

Closes #183


situx:
In many requirements, you removed the namespace from the definition. For example, a label now shows "dimension" instead of "geo:dimension" as a property. I feel that is a bit misleading and the namespace should be there despite the link going to another part of the spec.

nicholascar:
I removed them to use Class & Property labels consistently throughout.

Are you worried about how `dimension` might be interpreted in places like [Req 13](https://raw.githack.com/opengeospatial/ogc-geosparql/spec-iris/spec-iris.html#_standard_properties_for_geogeometry)?

We _can_ add `geo:` in for things, if you really want it there through but I'll need some more convincing!

nicholascar:
@situx I've tried one thing to avoid having to add `geo:` or `geof:` to the requirements: just call out that the things are 'functions' better. See **Req 40** at  https://raw.githack.com/opengeospatial/ogc-geosparql/spec-iris/spec-iris.html#_non_topological_query_functions:

> Implementations shall support the functions distance, buffer,...

So it's clear that it's `geof:distance` and it would also be clear that it's `geof:asWKT`, as opposed to `geo:asWKT` in other requirements. Where properties are meant, the Requirements already mention them, e.g. [Req 18](https://raw.githack.com/opengeospatial/ogc-geosparql/spec-iris/spec-iris.html#_property_geoaswkt) which says "Implementations shall allow the RDF property as WKT...".

So I think as long as the requirements include "...the functions..." or "...the properties..." there is no ambiguity.

I much prefer this method of achieving ambiguity than to add in `geo:` or `geof: because then we would have to revert from labels for properties/functions like _as WKT_ to the slugs _asWKT_ in order to make it `geo:asWKT` or `geof:asWKT`.

Can you live with this if I ensure all Requirements are worded in this way?

situx:
@nicholascar 
No I think this is not a good idea because of two reasons:
1. If we are really pedantic, then you would create new GeoSPARQL 1.1 requirements simply by not adding the namespace to the requirement text. E.g. in GeoSPARQL 1.0 we have "Implementations shall allow the RDFS class geo:SpatialObject to be used
in SPARQL graph patterns." but in GeoSPARQL 1.1 we have "Implementations shall allow the RDFS class SpatialObject to be used
in SPARQL graph patterns." Strictly speaking that would force me to create new IRIs for those requirements I think.
2. We currently have namespaces for functions geof: and geo: for vocabularies. But who is to say that will stay this way in the future. Maybe we will have two function namespaces in future releases. For that reason I think it is better to state the namespace explicitly in requirements.

nicholascar:
@situx I think you are conflating machine-readable preciseness (namespaces etc.) with human language preciseness. There is no ambiguity as to what "Implementations shall allow the RDFS class Spatial Object to be used" means because we have both stated the kind of thing, RDFS class, and name of the particular one, Spatial Object, which, in the context of this specification is unique. So for a human reader, this is enough. Additionally, however, we have _also_ hyperlinked the 'Spatial Object' to its definitional clause, so it's even clearer: a human can click on the term and go to its definition.

These Requirements are human-readable, not machine-readable. For example, the example you give says "Implementations shall allow... ". A machine can't read that so any machine use of this Requirement clause will necessarily need to be parsed by a human.

So our job is to ensure that a _human_ can't get it wrong and, as long as we state the type of thing (because we have two types of things with the same names, functions & properties) and give an unambiguous name for it (i.e. unique within the given context of the written Spec) then we are fine. We add in the hyperlink for ease of use.

With all the namespace and so on that you suggest, we are half way to a semantic document but this serves no purpose: these Requirements, in the form they are in here, can be machine-identified (the IRIs) but not machine interpreted.

situx:
Ok, in that case, I agree with your modifications.
I was not sure to which extents texts of requirements are allowed to be modified to not count as a modification of the requirement itself.

nicholascar:
Yeah, it's not exactly 100% determinable whether a modification has actually broken a normative part of a Requirement _because they are human readable things_!

Still, if we've faithfully carried the meaning of the Requirements from GeoSPARQL 1.0 through to 1.1, ad added new ones, and added better identification mechanisms around them and made them easier to read and use in the Spec, we will have done well!

When it comes to implementations, I expect we will be able to refer to an identified Requirement by IRI, and then to quote the Requirement text for reference, but the implementation will still only be able to prove support for the Requirement by demonstration, i.e. that the implementation does what the Requirement says it should. There will be no way to _prove_ that the Requirement is met in some sort of algorithmic way!

OK, if I've convinced you, approve this so I can get @FransKnibbe / @dr-shorthair / @mperry455 to approve too and we can merge and refine.

situx (approved):

FransKnibbe:
It does not stand in the way of approval, but I do prefer to have class/property names written as` namespace:name`. So instead of "GeoSPARQL does not restrict the cardinality of the [has default geometry](http://schemas.opengis.net/geosparql/1.0/geosparql_vocab_all.rdf#hasDefaultGeometry) property", have "GeoSPARQL does not restrict the cardinality of the [`geo:hasDefaultGeometry`](http://schemas.opengis.net/geosparql/1.0/geosparql_vocab_all.rdf#hasDefaultGeometry) property", for the following reasons:

- It seems to be common practice (for example in W3C ontology documentation).
- It helps reading; the ontology term is recognisable as a bit of code. And very likely implementers will want to view it that way. It is the code perspective they will need for implementation of the standard.
- It is makes clear that it not a general hyperlink, but a very specific one: an ontology term. So without resolving hyperlink a visual distinction can be made between hyperlinks to ontology terms and a hyperlink to another document, for example.
- Some labels consist of multiple words. like "has default geometry". That makes it easier to read them as a continuation of the words in the sentence, which is not our intention.

dr-shorthair:
I agree with @FransKnibbe - the key is to ensure that people are very clear when the reference is to an ontology term. 
Showing it as a cURI makes this clear.

nicholascar:
OK fine, I still disagree - I like human-readable text here - but I'll change to using cURIs since clearly the majority of yuo/us all want it. Will put in commits to this PR in a bit (tonight).

nicholascar:
OK, please have a look now: https://raw.githack.com/opengeospatial/ogc-geosparql/spec-iris/1.1/spec/spec.html.

Most of the references are to the appropriate section in the Spec but use cURI form of, e.g. `geo:Feature`. I'll keep adding a few more updates as there are still more links to be found and handled.

I've also added examples of `FeatureCollection` and `GeometryCollection` that were missing and done a lot of headings tidy-up (in Annex C, for example) that I've found that needed attention while working on all the links.

1.1/spec/09-Part-06.adoc
```diff
@@ -113,36 +118,48 @@ The class http://www.opengis.net/ont/geosparql#FeatureCollection[`geo:FeatureCol
...
+| *Req 6* Implementations shall allow the RDFS class <<Class: geo:FeatureCollection, `geo:FeatureCollection`>> to be used in SPARQL graph patterns.
 |http://www.opengis.net/spec/geosparql/1.1/req/core/feature-collection-class[`http://www.opengis.net/spec/geosparql/1.1/req/core/feature-collection-class`]
 |===
 
 === Standard Properties for geo:SpatialObject
 
-_To be defined_
+Spatial Object is essentially an abstract class and one that is not expected to be instantiated directly but rather used through more concrete subclasses of it. For this reason, no standard properties of Spatial Object are given here, directly. 
+
+Since GeoSPARQL defines two subclasses of Spatial Object, <<Class: geo:Feature, Feature>> & <<Class: geo:Geometry, Geometry>>, the union of their properties may be indirectly considered to be the standard properties of Spatial Object.
```
FransKnibbe, line 146: What is the purpose of this comment? I don't think the union of the properties of subclasses of SpatialObject makes much sense. 

Why not list the topology properties for SpatialObjects (e.g. geo:ehEquals) here?

FransKnibbe: The part "... two subclasses of Spatial Object, Feature & Geometry ... " can be read in several ways. How about "... two subclasses of Spatial Object: Feature & Geometry ... " or "... two subclasses of Spatial Object (Feature & Geometry) ... "?

1.1/spec/09-Part-06.adoc
```diff
@@ -113,36 +118,48 @@ The class http://www.opengis.net/ont/geosparql#FeatureCollection[`geo:FeatureCol
...
 
 === Standard Properties for geo:Feature
 
 Properties are defined for associating geometries with features.
```
FransKnibbe, line 150: Not really part of the topic of this PR, but this line no longer covers all properties.

1.1/spec/09-Part-06.adoc
```diff
@@ -183,56 +202,63 @@ SELECT (COUNT(*) AS ?cnt)
 WHERE { :f1 geo:sfDisjoint :f1 }
 ```
 
-Such cases are application-specific data modeling errors and are therefore outside of the scope of the GeoSPARQL specification., however it is recommended that multiple geometries indicated with http://www.opengis.net/ont/geosparql#hasDefaultGeometry[`geo:hasDefaultGeometry`] should be differentiated by `Geometry` class properties, perhaps relating to precision, SRS etc.
+Such cases are application-specific data modeling errors and are therefore outside of the scope of the GeoSPARQL specification., however it is recommended that multiple geometries indicated with <<Property: geo:hasDefaultGeometry, `geo:hasDefaultGeometry`>> should be differentiated by `Geometry` class properties, perhaps relating to precision, SRS etc.
```
FransKnibbe, line 205: Here we have the word Geometry formatted as code. All in all, many different formatting styles are used. I think it would be better not to use this type of formatting for ontology terms. So I suggest changing `Geometry` to <<Class: geo:Geometry, `geo:Geometry`>>

FransKnibbe:
Lots of changes for the better in this PR. Still, I notice many ways of formatting references to ontology terms:

1. As a hyperlinked cURI formatted as code (e.g. [`geo:SpatialObject`](http://www.example.com))
2. A hyperlink using the skos:label (e.g. [Spatial Object](http://www.example.com))
3. Capitalised plain text (e.g. "Feature")
4. Non-capitalised plain text (e.g. "feature")
5. Non-hyperlinked code (e.g. `Feature`)

These different styles can be confusing for the reader. I would like to suggest not to use the last two styles. But that is out of scope for this PR I think.

@nicholascar, I approve the PR if you'll take a look at the comments. And probably another PR is needed to make text formatting more consistent.

mperry455 (approved):

nicholascar:
> These different styles can be confusing for the reader. 

Yes, too many!

> I would like to suggest not to use the last two styles. 

Agreed

> But that is out of scope for this PR I think.

OK, I will create a follow-up PR to tidy up

> I approve the PR if you'll take a look at the comments. 

Great, I will go through the comments tonight.

> And probably another PR is needed to make text formatting more consistent.

No worries, I will, as mentioned above!

PR #193, opened 2021-08-12
Simple Terms & Defs

nicholascar:
@FransKnibbe I've added a few basic definitions for Semantic Web things and have left the "Spatial" section empty for you to add to!

I've also added a couple of missing Normative References, RDF for instance.

Rendered version of Spec with Ts & Ds: https://raw.githack.com/opengeospatial/ogc-geosparql/simple-td/1.1/spec/spec-td.html

### DIFF
--- a/1.1/spec/02-Introduction.adoc
+++ b/1.1/spec/02-Introduction.adoc
@@ -17,8 +17,8 @@ Subjects and objects within an RDF triple are called nodes and can also be be re
 Note that the same node may be a subject in some triples, and tan object in others.
 
 [#img-rdf]
-.RDF Triple
-image::img/01.png[600,400,align="center"]
+.RDF Triple  
+image::img/01.png[RDF Triple,600,400,align="center"]
 
 Almost all data can be presented or represented in RDF. In particular, it is an easy match to the (feature-instance-by-id, attribute, value) tuples of the General Feature Model <<ISO19109>>, and for the relational model as (table primary key, column, value).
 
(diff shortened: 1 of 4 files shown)


FransKnibbe:
Thank you @nicholascar. I notice the terms you have added to section 4 are not hyperlinked. Do we still want to pursue having a SKOS vocabulary for terms and definitions ([PR 165](https://github.com/opengeospatial/ogc-geosparql/pull/165)) and use it to link definitions to web resources? Admittedly, it is easier to just have a plain glossary (for now).

nicholascar:
> Do we still want to pursue having a SKOS vocabulary for terms and definitions

Not for the moment. It was too hard to do much other than create a flat list. Any hierarchy of Concepts in there became a problem due to the difficult of positioning these very diversely defined things.

So, let's just make a normal T&Ds section for now - just your spatial terms to be added - and then we can revisit whether there's benefit in creating a flat SKOS vocab of those terms. There may be - it may appear as yet another OGC vocab - but we just shouldn't try and redefined the whole world in it!

> Admittedly, it is easier to just have a plain glossary (for now).

Yes, let's do that for now.

situx (approved):
The definitions like that look good to me. I currently cannot think of essential definitions that are missing.

nicholascar:
OK, taking the contribution from @FransKnibbe as the same as approval, we have a second approval from @situx so let's merge this now and then follow up with any edits or additions, as needed.

Issue #103, opened 2021-03-06
rename SpatialMeasure?


FransKnibbe:
[PR 72](https://github.com/opengeospatial/ogc-geosparql/pull/72) introduced the term `SpatialMeasure`. I wonder if that is the best name for the concept. Firstly, the adjective "Spatial" should generally not be needed because the ontology is about spatial things. Secondly, we can not be certain that GeoSPARQL will never include semantics for Linear Referencing (LR). In LR, "measure" is an important term with an established meaning. So having two different kinds of measures could be confusing.
The overarching intent of the concept is to get an idea of the size of something, right? So how about "size" as a new name? Or perhaps "magnitude"?

dr-shorthair:
@FransKnibbe I was responsible for introducing it in the GEOX extensions we have been experimenting with at CSIRO. 
Yes, it is primarily just a scaled quantity (i.e. number with a unit of measurement). 
The 'spatial' part was to allow for the fact that some quantities are affected by the CRS in use. 
I don't understand your argument about linear referencing - that is still a spatial reference system after all. 

Nevertheless, I'd be comfortable with simplifying this concept. In fact - as I noted over in https://github.com/opengeospatial/ogc-geosparql/issues/83#issuecomment-780944827 it would probably be better implemented as a datatype. The `cdt:ucum` datatype implemented by Antoine Zimmerman and Maxime Lefrancois is nicely specified, though I'd like to see it moved into a more standard namespace.

FransKnibbe:
> @FransKnibbe I was responsible for introducing it in the GEOX extensions we have been experimenting with at CSIRO.
> Yes, it is primarily just a scaled quantity (i.e. number with a unit of measurement).
> The 'spatial' part was to allow for the fact that some quantities are affected by the CRS in use.
How do you mean that exactly? In the sense that distance can be measured in a plane or on a spherical surface?
If some quantities are differently affected by the CRS, perhaps it would be good to elaborate on that in the descriptions.

I hope there is a chance we can do without a unit of measurement altogether (and always use the unit of the CRS), but I raised that point already in [issue 98](98).

> I don't understand your argument about linear referencing - that is still a spatial reference system after all.

[...]

Issue #194, opened 2021-08-12
Which vocabularies for units of measurement does the OGC recommend?


FransKnibbe:
The documentation of SpatialMeasure says: 

> Properties for Spatial Measure may need to indicate a use a unit of measure. When they do, OGC recommended units of measure vocabularies should be used. See the OGC Definitions Server.

I cannot find a recommendation at the OGC Definitions Server. If there is such a recommendation, can we have a persistent link to it, instead of a general link to the Definitions Server?

ghobona:
On 2021-06-17, the OGC-NA discussed a proposal to deprecate OGC Units of Measure (UOM) URIs and to use QUDT.org URIs in GeoSPARQL.

The background to the discussion is at https://github.com/opengeospatial/NamingAuthority/issues/80#issuecomment-863587536

OGC UOM URIs are those that have a base http://www.opengis.net/def/uom


The OGC-NA decided to delegate the issue of which UOM URIs to use in GeoSPARQL to the GeoSPARQL SWG to decide. 

The GeoSPARQL SWG's decision will not bind other SWGs that are already using OGC UOM URIs.

FransKnibbe:
Thank you for shedding light on the matter, @ghobona. It does leave me confused. So the GeoSPARQL SWG is to pick a single UOM vocabulary? I was not aware of that. But I do think that having too much freedom of choice for UoM can hamper data interoperability.

Since the reason to have liberty of choice in UoM is to be able to use distance units that are not possible to convert to metric, probably the SWG should recommend the vocabulary that is best at expressing historical units of distance.

dr-shorthair:
1. Going back to the last century/millennium, OGC recommended UCUM for unit symbols or codes (in the WMS spec). 

2. If you prefer something that uses URIs, it is hard to go past QUDT. 

AFAIK those are the most widely known and used general-purpose vocabularies or systems for units-of-measurement. 
When OGC (and EPSG) started developing its own URI-based system, QUDT was not available. 
Now that it is, this makes one more thing that OGC does not have to manage itself.

ghobona:
@FransKnibbe The OGC-NA did not place restrictions on the number of UoM vocabularies that the GeoSPARQL SWG may use in the new version of GeoSPARQL. It is for the GeoSPARQL SWG to select one or more that are appropriate for use in GeoSPARQL.

FransKnibbe:
[...]

Issue #178, opened 2021-07-28
Change domain of spatial scalar properties from Feature to SpatialObject


nicholascar:
As decided in teleconference 2012-07-29

situx:
A few questions concerning this:
Do we also need to change the requirements for properties of geo:Geometry to include these?
Besides, in the spec there are two sections:
Standard properties for geo:SpatialObject and
Standard properties for geo:SpatialMeasure

Both sections have the content "To be defined"
What is supposed to be added there?

FransKnibbe:
I kind of forgot about [this earlier comment](https://github.com/opengeospatial/ogc-geosparql/issues/33#issuecomment-825552469). But it still applies, I think. It makes little sense to let measurements of size partake in topological relationships. They are missing a frame of reference that Geometries and Features do have.

nicholascar:
> Do we also need to change the requirements for properties of geo:Geometry to include these?

Yes, and I've done that in the relevant PR.

> Besides, in the spec there are two sections: Standard properties for geo:SpatialObject and Standard properties for geo:SpatialMeasure. Both sections have the content "To be defined" What is supposed to be added there?

I have added notes to the section on `geo:SpatialMeasure` to say that essentially GeoSPARQL 1.1 doesn't make any recommendations about properties for it normatively, however there are a number of examples (informative) that show how it _could_ be used with properties, for example `qudt:numericValue` & `qudt:unit`.

I've added, in PR #186, a couple of sentences for `geo:SpatialObject` that basically say it's abstract and that `Feature` & `Geometry`'s stnadar properties are its standard props.

nicholascar:
> I kind of forgot about this earlier comment. But it still applies, I think. ...

The remaining question there, about relating Spatial Measure to QUDT's  `qudt:Quantity` ,or similar, is picked up in the PR #195, so let's carry on discussion there as PR #184 will likely soon close this Issue.

nicholascar:
The SpatialMeasure class is now gone, so this Issue has no more work to be done.

PR #195, opened 2021-08-12
Remove SpatialMeasure

FransKnibbe:
~~A straightforward set of changes: the class geo:SpatialMeasure is renamed to geo:SpatialSize.~~ 
After some discussion (in this PR) it was decided to remove geo:SpatialMeasure altogether, and use only properties for specifying size.  Resolves [issue 103](https://github.com/opengeospatial/ogc-geosparql/issues/103).


nicholascar:
I'm not overly thrilled by the name "Spatial Size" but I don't fully know why! 

I think it may have something to do with the sense that "size" is a property of something (in various forms as length, volume etc) but that "measure" is not a property but a view of a property. So "Spatial Measure" then is a view of, a measure of, something spatial, not an actual spatial thing itself, a size.

I could easily have this all wrong though and that a measure of something might also be a property of it.

One alternative that has occurred to me, and this is mainly for @dr-shorthair to comment on, is "Spatial Property". Why is "Spatial Measure" the class name not "Spatial Property" when elsewhere we have "Observable Property", "Statistical Property" etc.? If "Spatial Measure" is really just a spatial "Observable Property", why not call it "Spatial Property"?

@FransKnibbe: would you be ok with "Spatial Property" instead of "Spatial Size" or is this missing your intention?

FransKnibbe:
Frankly  am not too crazy about ¨SpatialSize¨ either. Partly because of the ¨Spatial¨ prefix; GeoSPARQL is about spatial data, so we could use that prefix everywhere. Or not, which would be preferable - ¨Spatial¨ seems superfluous to me. Then we would be left with ¨Size¨ only. I do maintain that the concept of size does perfectly cover the intention and scope. Still the name would be funny because it is the name of a class, while conceptually it is a (super)property. Modelling as a class leads to thinking about this as being able to be in the subject position of an RDF triple, and being able to speak of ¨a Size¨. Which could be considered slightly jarring.

I think modelling a size indication as a class contributes to it being hard to find a good name for the class. The suggestion to rename ¨SpatialMeasure¨ to ¨SpatialProperty¨ reflects that, I think. It is more a property than a class. But ¨Spatial Property¨ is too broad a term, I think. All properties defined in GeoSPARQL could be considered spatial properties. And to name a class a property can also lead to confusion.

Actually, in hindsight I am trying to follow why SpatialMeasure is set up as a class. An indication of the size of a Feature is much easier thought of as a property. Do I understand correctly that it became a class because it was not possible to choose between modelling it as a datatype property or an object property? Then [Issue 194 ](https://github.com/opengeospatial/ogc-geosparql/issues/194 )comes into play: If the UoM vocabulary prescribed only allows complex descriptions of quantities and units, then why not model Size/SpatialMeasure as an ObjectProperty?

If it were to become an ObjectProperty, and we do not want people to force using blank nodes, we do need to find a good general term for the concept of an observation (a measurement or estimate of size). Is [om-lite](http://def.seegrid.csiro.au/ontology/om/om-lite) usable for that? I can´t tell, because currently it is offline. 

The longer I think about it, the more I get the idea the basic problem is the way size is modelled. It being a property would be much more fitting. Other problems (e.g. [issue 178](https://github.com/opengeospatial/ogc-geosparql/issues/178)) would disappear too. 

Going back to the naming of the beast, how about DeterminationOfSize (if it is a class) or hasDeterminationOfSize (if it is a superproperty)?

dr-shorthair:
> Actually, in hindsight I am trying to follow why SpatialMeasure is set up as a class. An indication of the size of a Feature is much easier thought of as a property. Do I understand correctly that it became a class because it was not possible to choose between modelling it as a datatype property or an object property? Then Issue 194 comes into play: If the UoM vocabulary prescribed only allows complex descriptions of quantities and units, then why not model Size/SpatialMeasure as an ObjectProperty?

It was carried over from GeoSPARQL Extensions ontology that was developed for Loc-I. The general principle was that size strictly requires multiple pieces of information - a number, a scale (unit-of-measurement), potentially a CRS, maybe some quality measures - so a class was needed to support multiple properties. 

However, on further investigation and reflection, I'm now of the opinion that only number and scale are essential (the other properties can be moved elsewhere). Furthermore, there are community expectations about how a scaled quantity should be expressed as a literal with the pattern `1500 m` or `110 km/hr'. 

While the RDF representation does not have to match the natural language representation, there is a nice implementation of this pattern at [cdt:ucum[(https://w3id.org/cdt/ucum) data-type developed by @maximelefrancois86 and Antoine Zimmerman. I'm inclined to adopt this for scalar quantities. It would simplify the GeoSPARQL ontology, though of course it does require some additions to the processing side (Java is already available).

nicholascar:
> I'm now of the opinion that only number and scale are essential

Perhaps (I can't think of other necessary things), so let's say yes.

> there is a nice implementation of this pattern in the cdt:ucum data-type developed by @maximelefrancois86 and Antoine Zimmerman. I'm inclined to adopt this for scalar quantities. It would simplify the GeoSPARQL ontology, though of course it does require some additions to the processing side (Java is already available).

'nice' is extremely subjective. While I see some benefits to this, it imposes a huge burden on implementers - the whole set of UCUM parsing algorithms - whereas a number/unit split doesn't. With a number/unit split, you can, at the very least, know whether your values are commensurate or not, even if you can't translate between two that aren't.

@dr-shorthair you have said in the past that you do care what implementers think and I'm an implementer here (RDFlib) and I think that UCUM is hard, too hard for just now at least.

Also, whatever the benefits of UCUM, I don't think it's timely for this. We need to complete this GeoSPARQL 1.1 ASAP (we are currently a month or two behind where we said to everyone we would be).

@FransKnibbe:

> Then we would be left with ¨Size¨ only..

Actually I'm OK with this!

> Modelling as a class leads to thinking about this as being able to be in the subject position of an RDF triple, and being able to speak of ¨a Size¨. Which could be considered slightly jarring.

On reflection:

```
ex:A geo:hasArea ex:B .

ex:B
    a geo:Size ;  # "Represents a measurement of some dimension of a Spatial Objects's spatial presence."
    qudt:numericValue 250 ;
    qudt:unit qudt:M2 ;
.
```

or

```
ex:A geo:hasArea ex:B .

ex:B
    a geo:Size ;  # "Represents a measurement of some dimension of a Spatial Objects's spatial presence."
    ucum:value "250 m^2" ;  # or whatever
.
```

In both cases, it's fine to for the `Size` instance "...being able to be in the subject position of an RDF triple...".

### Recommendations
1. We go with renaming `SpatialMeasure` -> `Size`
2. We say little about "standard properties" for `Size` so people can do whatever, QUDT, UCUM etc. 
    * we've already done this, so no change needed
3. We add alternate (to QUDT) examples in the Annexes for UCUM use with `Size`
    * @dr-shorthair this is for you!
4. We say something more about mapping `Size` to `sosa:ObservableProperty`
    * I do think it's just `geo:Size rdfs:subClassOf sosa:ObservableProperty` with a note about `Size` being a spatial OP
    * There is no dimensional classifier in SOSA for "spatial" but there is in QUDT, so we could, informally, add a dimension vector for `Size` using QUDT

dr-shorthair:
> Also, whatever the benefits of UCUM, I don't think it's timely for this. We need to complete this GeoSPARQL 1.1 ASAP (we are currently a month or two behind where we said to everyone we would be).

Agreed. 

> or
>```
>ex:A geo:hasArea ex:B .
>
>ex:B
>    a geo:Size ;  # "Represents a measurement of some dimension of a Spatial Objects's spatial presence."
>    ucum:value "250 m^2" ;  # or whatever
>.


I'd suggest 
```
ex:A geo:hasArea ex:B .

ex:B
    a geo:Size ;  # "Represents a measurement of some dimension of a Spatial Objects's spatial presence."
    rdf:value "250 m2"^^cdt:ucum ;  
.
```
or better 
```
ex:A geo:hasArea "250 m2"^^cdt:ucum .

**Comment by nicholascar** [comment]
> I'd suggest
> or better

OK, I'll put in some examples of each.

Seems like you're OK with `Size` as the class name. @FransKnibbe, do you want to convert to `Size` with @dr-shorthair  and I supporting the change?

**Comment by FransKnibbe** [comment]
> However, on further investigation and reflection, I'm now of the opinion that only number and scale are essential (the other properties can be moved elsewhere). Furthermore, there are community expectations about how a scaled quantity should be expressed as a literal with the pattern `1500 m` or `110 km/hr`.
> 
> While the RDF representation does not have to match the natural language representation, there is a nice implementation of this pattern in the [cdt:ucum](https://w3id.org/cdt/ucum) data-type developed by @maximelefrancois86 and Antoine Zimmerman. I'm inclined to adopt this for scalar quantities. It would simplify the GeoSPARQL ontology, though of course it does require some additions to the processing side (Java is already available).

The [OM vocabulary](http://www.foodvoc.org/) also seems to support having the number and unit in the same literal. Of course software/library support is another matter.

**Comment by nicholascar** [comment]
> The OM vocabulary also...

Well please do add in OM examples for the use of `Size`! 

But @FransKnibbe what do you think about the 4 Recommendations above?

**Comment by dr-shorthair** [comment]
OM is pretty good, but not as widely used as QUDT or UCUM, and governance is not as solid. It is dependant on one person (the originator).

**Comment by FransKnibbe** [comment]
> Seems like you're OK with `Size` as the class name. @FransKnibbe, do you want to convert to `Size` with @dr-shorthair and I supporting the change?

I still see some remaining issues that have been mentioned before but are scattered over time and separate issues. With this thead currently in focus, let me repeat them:

1. Why is SpatialMesure modelled as a class and not an ObjectProperty?
2. SpatialMeasure as a class has problems with finding the right superclass. Currently it is thought of as a subclass of SpatialObject. But SpatalObjects can have topological relationships, which does not make sense for SpatialMeasure.
3. If we agree with the assessment of @dr-shorthair that only number and unit are essential, then why not fix the units to metre and use only DatatypeProperties (like the metric size properties we already have)? The only remaining reason I can think of is that we do not want to exclude historic units of distance that do not have a known conversion factor to metres.  [Yojana](https://en.wikipedia.org/wiki/Yojana) for example. But as far as I can tell, it is not possible to use Yojana with QUDT or UCUM. 

So an alternative suggestion would be: Remove SpatialMeasure altogether and leave only the metric size indication properties. It would make things much simpler (for implementers, data publishers and data consumers) and data interoperability would be helped a lot.  What would be lost in that case?

**Comment by situx** [comment]
In my opinion, we should at least offer the possibility for users to define spatial measures which are not defined in the metric system. I could think of many use cases in which the metric representation is irrelevant, e.g. in the historical sciences as Frans pointed out. But maybe it is simply a necessity of the data provider to provide sizes e.g. not in kilometers but in miles. If we only define metric properties, then the burden of returning those measures in miles and not in kilometers is on the triple store implementation, instead of simply defining the relevant size in its relevant unit.

**Comment by FransKnibbe** [comment]
> In my opinion, we should at least offer the possibility for users to define spatial measures which are not defined in the metric system.

I think using non-metric distance units that have a known conversion factor should not be encouraged in any way. The SI exists to have an end to people using different units for the same thing. The WWW is precisely the place were this kind of standardisation is needed.

Still, we could offer something for cases where extra complexity is really unavoidable. For example an ObjectProperty like hasSize (with appropriate subporperties like hasLength etc.), without a range specification, so that people can use whatever they want. But it should come with a warning that data will have poor interoperability.

**Comment by situx** [comment]
Yes, I think that is a good solution if we were to switch to properties.
I totally agree that wherever possible one should use the metric units and we should encourage their usage.
A warning of poor interoperability is a good way of making that clear, while at the same time not excluding the possibility of a definition in other units.

**Comment by nicholascar** [comment]
> 1. Why is SpatialMesure modelled as a class and not an ObjectProperty?

Due to following the [SOSA Observable Property](https://www.w3.org/TR/vocab-ssn/#SOSAObservableProperty) model.

> 2. SpatialMeasure as a class has problems with finding the right superclass. Currently it is thought of as a subclass of SpatialObject. But SpatalObjects can have topological relationships, which does not make sense for SpatialMeasure.

It's not actually. [In our ontology, it has no superclass.](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/geo.ttl#L689).

> 3. If we agree with the assessment of @dr-shorthair that only number and unit are essential, then why not fix the units to metre and use only DatatypeProperties (like the metric size properties we already have)? The only remaining reason I can think of is that we do not want to exclude historic units of distance that do not have a known conversion factor to metres. Yojana for example. But as far as I can tell, it is not possible to use Yojana with QUDT or UCUM.

> If we agree with the assessment of @dr-shorthair that only number and unit are essential

I think we all do, for simple scalar values.

> then why not fix the units to metre and use only DatatypeProperties

As per the original conversation about the inclusion of metric properties: I think it's great to provide the metric datatype properties but I don't think we can exclude non-metric ways of conveying area. This would be akin to saying the only GeoSPARQL way of recording geometry is using, say, GML and you're not allowed to use KML, WKT etc. GeoSPARQL should provide options. 

I really like the idea of providing `hasMetricArea` etc and I expect to use that form of the properties often, however when GeoSPARQL's used for nanometer objects, achres, etc. I don't want to have to force users to either convert data or implement other properties, which is what they have been doing to date.

Fundamentally, this ontology should provide good data representative power and let users work out things like conversion. This is an ontology, not a calculation library! But with `hasArea` and `hasMetricArea`, we have potentially all options...

> as far as I can tell, it is not possible to use Yojana with QUDT or UCUM.

Systems like QUDT are continuously growing to accomodate things, so could both include Yojana and other non-convertable things. Let's leave conversions to the UoM systems and just cater for their use.

> So an alternative suggestion would be: Remove SpatialMeasure altogether and leave only the metric size indication properties. It would make things much simpler (for implementers, data publishers and data consumers) and data interoperability would be helped a lot. What would be lost in that case?

Sorry, I just can't agree to this, for the reasons given above! But you've suggested a way out:

> Frans: we could offer something for cases where extra complexity is really unavoidable. For example an ObjectProperty like hasSize (with appropriate subproperties like hasLength etc.), without a range specification, so that people can use whatever they want. But it should come with a warning that data published that way will have poor interoperability.

> Frans: This approach would also eliminate the need for the SWG to recommend one or more UoM vocabularies (see issue 194).

> Timo: Yes, I think that is a good solution if we were to switch to properties.

> Timo: A warning of poor interoperability is a good way of making that clear, while at the same time not excluding the possibility of a definition in other units.

I'm ok with this! Let's do introduce `geo:hasSize` & `geo:hasMetricSize` as superproperties of the others. Then, like we've done with the specific AusPIX DGGS, we can include non-normative examples of QUDT, OM & UCUM use. So your options are supply a literal value in metres for the _metric_ properties, or a UCUM literal, whcih includes units, or a Blank Node of with value & units for QUDT etc. Conversions from things like QUDT to metric literals are left to implementers to work out from whatever UoM ontology they are using.

I'm happy to draft a PR for this if we are agreed on this direction.

**Comment by FransKnibbe** [comment]
> > 1. SpatialMeasure as a class has problems with finding the right superclass. Currently it is thought of as a subclass of SpatialObject. But SpatalObjects can have topological relationships, which does not make sense for SpatialMeasure.
> 
> It's not actually. [In our ontology, it has no superclass.](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/geo.ttl#L689).

That´s right. I was confused.

> As per the original conversation about the inclusion of metric properties: I think it's great to provide the metric datatype properties but I don't think we can exclude non-metric ways of conveying area. This would be akin to saying the only GeoSPARQL way of recording geometry is using, say, GML and you're not allowed to use KML, WKT etc. GeoSPARQL should provide options.

It is possible to make that comparison, but I don think it goes all the way. Metre is a much more powerful standard than WKT, GML, GeoJSON, etc. My thinking is that we have to live with those different formats temporarily, because there is no good general alternative. Having so many different fomats is not a good thing by itself.

> I really like the idea of providing `hasMetricArea` etc and I expect to use that form of the properties often, however when GeoSPARQL's used for nanometer objects, achres, etc. I don't want to have to force users to either convert data or implement other properties, which is what they have been doing to date.

Well, nanometres are metric. And for the sake of data interoperability it _would_ be good to encourage people not to use acres (I see there are several kinds of them too). 

> Fundamentally, this ontology should provide good data representative power and let users work out things like conversion. This is an ontology, not a calculation library! But with `hasArea` and `hasMetricArea`, we have potentially all options...

Another point of view is saying that any standard necessarily limits options. But of course freedom of expression should not suffer from that.

> > as far as I can tell, it is not possible to use Yojana with QUDT or UCUM.
> 
> Systems like QUDT are continuously growing to accomodate things, so could both include Yojana and other non-convertable things. Let's leave conversions to the UoM systems and just cater for their use.

Right. And because it is difficult to predict how UoM systems will develop I think it good not to link any single one directly to GeoSPARQL.

> I'm ok with this! Let's do introduce `geo:hasSize` & `geo:hasMetricSize` as superproperties of the others. Then, like we've done with the specific AusPIX DGGS, we can include non-normative examples of QUDT, OM & UCUM use. So your options are supply a literal value in metres for the _metric_ properties, or a UCUM literal, whcih includes units, or a Blank Node of with value & units for QUDT etc. Conversions from things like QUDT to metric literals are left to implementers to work out from whatever UoM ontology they are using.
> 
> I'm happy to draft a PR for this if we are agreed on this direction. I'm also happy for one of you to do that! Such a PR should close this one.

Would that include removing the class SpatialMeasure?
Anyway, I think it would be good to have the superproperties. At least it avoids a lot of duplicates of explanations. Which would only grow, because we could see demand for additional size indications (width, depth, height, thickness, circumference, radius, ...).

As far as I can tell there are different scenario´s for wanting to be able to use more complex size indications:
1. To add quality metadata to a measurement or estimate
2. To be able to use non-metric units with a UoM vocabulary
3. To be able to use historical non-metric units (that may not have a clear conversion factor and may not be possible to express using a UoM vocabulary)

To help these underway, we could provide examples for all.

I would be happy to help with a new PR.

**Comment by dr-shorthair** [comment]
This discussion is important. 

Is the goal of the ontology to provide a way to capture the current representation provided by the data supplier, or to provide a standard representation for data consumers. 

My earlier experience with (for example) GML was that we went out of our way to make it easy for providers, and allowed for many options. Too many for consumers, so the generic GML client turned out to be impossible to build. 

_Someone_ has to do the work of transformation (here: from any length scale --> metres). Should that be the provider or the consumer? In the world we are trying to move towards there would be many more consumers than providers, so I think the onus should be on the provider to convert their data to the standard (including the standard length scales) so that client applications are easy to build.  

So overall I agree with @FransKnibbe proposition here. Too many options is _not_ a standard.

**Comment by situx** [comment]
Is the handling of units not something that should be done in a triplestore implementation?
In the same way as there are software libraries for converting between different crs such libraries for unit conversion and unit awareness would just need to be part of a triple store implementation and possibly supported in SPARQL extension functions. 
Am I wrong here? 
If I am not then I think the solution of providing an option for entering non metric units makes sense but a recommended way to define metric units is good as well

**Comment by FransKnibbe** [comment]
> Is the handling of units not something that should be done in a triplestore implementation?

Despite the name, GeoSPARQL can be useful outside of triple stores with SPARQL engines (see also[ issue 156](https://github.com/opengeospatial/ogc-geosparql/issues/156)). For example, spatial data modelled with GeoSPARQL could be file-based or exist as microdata on a web page.

Besides, even if something is feasible with behind-the-scenes software libraries, that doesn't make it recommendable. If I try to imagine how a federated SPARQL query comparing the sizes of two spatial objects would look if the sizes are encoded with arbritrary units from arbritrary UoM vocabularies, I see a lot of complexity that may not be necessary.

Still, I think use cases that do need extra complexity when compared to using metric values are valid. Particularly the ability to use historical distance units could be important for e.g. historical or archeological data. But I think it would be good for GeoSPARQL to recommend using metric values where possible, for the sake of simplicity and interoperability.

So yes, I agree with
> I think the solution of providing an option for entering non metric units makes sense but a recommended way to define metric units is good as well

**Comment by nicholascar** [comment]
> I think the solution of providing an option for entering non metric units makes sense but a recommended way to define metric units is good as well

I also agree

> Would that include removing the class SpatialMeasure?

Yes. If our recommended way to convey size is to use metric data properties and our other way is to use value/unit pairs or micro format value/unit literals, we just supply the properties and leave the range value to the users, prompted by examples. No need for a class.

**Comment by FransKnibbe** [comment]
> Please follow the easiest path, so if you'd like to adapt this one, do that!

OK, I will try to repurpose this PR. Let's see how that goes.

**Comment by FransKnibbe** [comment]
@dr-shorthair, @nicholascar  and @situx: I could use some help with an example for Annex B using a historic unit of length that has no conversion factor to metre. The example I am trying to express uses the unit http://dbpedia.org/resource/Schoenus (the route from Seleucia to Artemita is 15 Schoeni, according to a statement made some 2 kiloyear ago by http://dbpedia.org/resource/Isidore_of_Charax). So I am looking for something like a hasUnit property that I can use to encode the size estimate. Does someone have a suggestion? Qudt:hasUnit has range qudt:Unit, so I can't use that.
The [OM vocabulary](http://www.foodvoc.org/) is offline at the moment. That doesn't help.

**Comment by situx** [comment]
Hi Frans, just a heads up that you can find OM on Github https://github.com/HajoRijgersberg/OM

**Comment by FransKnibbe** [comment]
Thanks @situx! Unfortunately it seems I can't use om:hasUnit either.

**Comment by dr-shorthair** [comment]
There are multiple web-based units-of-measure systems. There is also current work in OBO underway. Not clear which one will achieve the most widespread adoption. I would say that only UCUM has a real organizational backing but it is not really web-based! I think we need to keep our options open here. So I would not adopt a single predicate from any of these systems, also because dependencies will bring in a while lot of other baggage.

**Comment by dr-shorthair** [comment]
> Qudt:hasUnit has range qudt:Unit, so I can't use that.

Why not? It merely entails that the object can be understood to be classifiable as a qudt:Unit (as well as any other classification already applied to it).

**Comment by FransKnibbe** [comment]
OK, here is an attempt at an example for using geo:hasLength with a historical unit of distance:

```
@prefix dbp: <http://dbpedia.org/resource/> .
@prefix qudt: <http://qudt.org/schema/qudt/> .
ex:Seleucia_Artemita
    a geo:Feature ;
    skos:prefLabel "The route from Seleucia to Artemita."@en ;
    geo:hasLength [
      qudt:unit ex:Schoenus ;
      qudt:value "15"^^xsd:integer ;
    ]
.
ex:Schoenus
  a qudt:Unit;
  skos:exactMatch dbp:Schoenus;
.
```

Is this a good example?

FransKnibbe:
I have reworked quite a bit, I hope I caught everything that needed to change.

situx (approved):
Looking good. I saw a few typos here and there, but nothing that would discourage me from approving.

nicholascar (approved):

PR #205, opened 2021-09-02
Bibtex citation

situx:
This pull request includes bibtex files which I would like to include
* In the repository
* Render on the Github page

I find it very important that we communicate how to cite the standard properly.
If you look at Google Scholar for GeoSPARQL standard, you will see that there is a variety of options to choose from, not all of them great. So we should communicate how we want to be cited.

I would encourage everyone to edit these bibtex files until we all agree.

### DIFF
--- a/1.0/citeme.bib
+++ b/1.0/citeme.bib
@@ -1,5 +1,5 @@
 @techreport{perry2012ogc,
-  title={OGC GeoSPARQL-A geographic query language for RDF data},
+  title={OGC GeoSPARQL - A geographic query language for RDF data},
   author={Perry, Matthew and Herring, John},
   journal={OGC implementation standard},
   volume={40},
--- a/1.0/citeme.bib
+++ b/1.0/citeme.bib
@@ -1,0 +1,7 @@
+@techreport{perry2012ogc,
+  title={OGC GeoSPARQL-A geographic query language for RDF data},
+  author={Perry, Matthew and Herring, John},
+  journal={OGC implementation standard},
+  volume={40},
+  year={2012}
+}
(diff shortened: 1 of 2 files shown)


oldskeptic:
"I find it very important that we communicate how to cite the standard properly." +1

dr-shorthair:
Indeed. Almost every time I review a journal paper, I have to tell them to cite standards in the bibliography, not just in footnotes. That is made a lot easier if there is a clear form of citation provided.

dr-shorthair:
a .cff file is another useful tool 
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files

situx:
@dr-shorthair I think what we are lacking at OGC is Cite Me Buttons in general. In IETF standards you can download a bibtex on their homepage.
W3C does not have something like that but there is this tool https://w2.syronex.com/jmr/w3c-biblio you could use to generate a bibtex citation.
We should have something like that for OGC standards

dr-shorthair:
Also: I've been nagging OGC secretariat for ca. 15 years about the value/need to get DOIs for OGC standards, at least. 
@ghobona could we revive this?

dr-shorthair:
(I suspect that research paper writers would take OGC docs more seriously if there were DOIs)

1.0/citeme.bib
```diff
@@ -0,0 +1,7 @@
+@techreport{perry2012ogc,
+  title={OGC GeoSPARQL-A geographic query language for RDF data},
```
nicholascar, line 2: please just format as per 1.1 with spaces around the '-'.

nicholascar (approved):

1.0/citeme.bib
situx: Done

PR #213, opened 2021-09-23
Added prefixes to aggregate functions

situx:
As discussed in the last call, prefixes have now been added to the aggregate functions


nicholascar (approved):
This is a straightforward find-and-replace PR that we've all agreed on, so I'm going to merge directly (i.e. without further reviews) to save us all time.

nicholascar:
@situx do we need examples of these functions' use in Annex B? I don't see any at the moment. In fact, we should check that we have examples for all functions. 

I've added lots for the classes and properties, but these will need a review about now due to new property additions etc, but not the functions.

situx:
Yes it would make sense to add examples. I have not added any examples for aggregate functions as of yet

PR #214, opened 2021-09-25
Editorial changes only

nicholascar:
Spec:
* formatting
* spelling
* removed "remove when fixed" notes
* section IRIs (in Annex B) to make simpler
* heading levels (in Annex A) to make uniform
* Added examples for properties missing them

Ont: 
* made isDefinedBy IRIs consistent
* updated examples links

FuncsRules:
* updated examples links

No normative changes.

See rendered result at https://raw.githack.com/opengeospatial/ogc-geosparql/editorial-only/1.1/spec/spec.html

PR #215, opened 2021-09-25
Tidy SO Properties

nicholascar:
The defining RDF in the Spec for the Spatial Object properties (hasSize etc) has been aligned with other defining RDF : formatting & prefixes.

The ontology SO Props have been aligned with the Spec's RDF.


nicholascar:
moved these commits to PR #214

PR #216, opened 2021-09-25
Add an ontology overview Figure

nicholascar:
and tidy the annotations of other figures.

See the new Figure rendered in place for review at https://raw.githack.com/opengeospatial/ogc-geosparql/ontology-diagrams/1.1/spec/spec-ont-fig.html#_core


mperry455:
Looks good Nick, but I believe those Simple Features, Egenhofer and RCC8 relations are defined on SpatialObject rather than Feature.

situx:
@nicholascar I think you could add the rdfs:subClassOf relations of the collection classes to make the diagram complete

nicholascar:
> Looks good Nick, but I believe those Simple Features, Egenhofer and RCC8 relations are defined on SpatialObject rather than Feature.

Oops, I've fixed that.

> I think you could add the rdfs:subClassOf relations of the collection classes to make the diagram complete.

I've added those.

nicholascar:
> I would put ogc:geomLiteral there because it is also used in the standard

Done

nicholascar:
OK, I think I've addressed the issues identified.

situx (approved):
Looking good now

mperry455 (approved):
Looks good to me.

PR #219, opened 2021-09-30
Fixes for SHACL shapes

situx:
For our GeoSPARQL benchmark I have tried to validate the benchmark dataset we use with the SHACL shapes.
You can find the dataset here: https://github.com/OpenLinkSoftware/GeoSPARQLBenchmark/blob/master/src/main/resources/gsb_dataset/dataset.rdf

I have changed the following things in the shapes:
- For all literal types: Check also for empty literals and consider whitespace at the beginning and end of regexes. Using pyshacl this does not seem to work for multiline literals though 

I have observed the following behaviors when applying the shapes:
- The benchmark dataset contains the self-defined relations my:hasExactGeometry and my:hasPointGeometry which are rdf:subPropertyOf geo:hasGeometry which cause a validation error when applying the SHACL shapes, but should not
- Shape S32-geometryClass-hasSerialization-sub forces only one geometry serialization for a Geometry and punishes this with a violation error, but S1-b-hasGeometry-hasSerialization-sub just gives a warning. Should we maybe change S32-geometryClass-hasSerialization-sub to check for at least one outgoing geo:hasSerialization?

Any ideas how to fix the last two issues? Can we only do this with reasoning enabled?
@nicholascar @mathib any ideas?

### DIFF
--- a/1.1/validator.ttl
+++ b/1.1/validator.ttl
@@ -607,7 +607,7 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however 
 <S16-wkt-content-sub-start>
 	a sh:PropertyShape ;
 	sh:path geo:asWKT ;
-	sh:pattern "^(P|C|S|L|T|<)" ;  # starts with P, C, S, L, T, or < for Point, TIN etc. or IRI
+	sh:pattern "^\\s*$|^\\s*(P|C|S|L|T|<)" ;  # starts with P, C, S, L, T, or < for Point, TIN etc. or IRI
 	sh:message "The content of an RDF literal with an incoming geo:asWKT relation must conform to a well-formed WKT string, as defined by its official specification (Simple Features Access)."@en ;
 	skos:example 
 		"""
@@ -641,7 +641,7 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however 
 <S17-gml-content-sub-start>
 	a sh:PropertyShape ;
 	sh:path geo:asGML ;
-	sh:pattern "^(<)(.+)(>)$" ;  # starts with < ends with >
+	sh:pattern "^\\s*$|^\\s*(<)(.+)(>)\\s*$" ;  # starts with < ends with >
 	sh:message "The content of an RDF literal with an incoming geo:asGML relation must conform to a well-formed GML geometry XML string, as defined by its official specification."@en ;
 .
 
@@ -654,7 +654,7 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however 
 <S18-geojson-content-sub-start>
 	a sh:PropertyShape ;
 	sh:path geo:asGeoJSON ;
-	sh:pattern "^({)(.*)(})$" ;  # starts with { ends with }
+	sh:pattern "^\\s*$|^\\s*({)(.*)(})\\s*$" ;  # starts with { ends with }
 	sh:message "The content of an RDF literal with an incoming geo:asGeoJSON relation must conform to a well-formed GeoJSON geometry string, as defined by its official specification."@en ;
 .
 
@@ -667,7 +667,7 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however 
 <S19-kml-content-sub-start>
 	a sh:PropertyShape ;
 	sh:path geo:asKML ;
-	sh:pattern "^(<)(.+)(>)$" ;  # starts with < ends with >
+	sh:pattern "^\\s*$|^\\s*(<)(.+)(>)\\s*$" ;  # starts with < ends with >
 	sh:message "The content of an RDF literal with an incoming geo:asKML relation must conform to a well-formed KML geometry XML string, as defined by its official specification."@en ;
 .
 


nicholascar:
> Any ideas how to fix the last two issues? Can we only do this with reasoning enabled?

We can't use reasoning with shape S32-geometryClass-hasSerialization-sub for this will guarentee it will fail. If a subproperty of hasSerialization exists and RDFS or greater reasoning is used (essentially any reasoning) then there will be two properties.

Is your first point related, i.e. do you only get a violation with reasoning turned on? Reasoning will infer that both geo:hasSerialization exists wherever my:hasExactGeometry or my:hasPointGeometry does, thus two properties.

situx:
I have attached two results. The first one is without reasoning, the second one is with reasoning.
[validationresult.txt](https://github.com/opengeospatial/ogc-geosparql/files/7271385/validationresult.txt)
[validatorresult2.txt](https://github.com/opengeospatial/ogc-geosparql/files/7271454/validatorresult2.txt)
I think with this it is easier to identify the issues

situx:
- [ ] Create Annex for SHACL usage
- [ ] Tidy up SHACL rules
- [ ] Shift some SHACL rules to the new community repository

mathib:
Hey Timo and Nick, I'll try to provide some inputs (and sorry for my unavailability of last months):

> The benchmark dataset contains the self-defined relations my:hasExactGeometry and my:hasPointGeometry which are rdf:subPropertyOf geo:hasGeometry which cause a validation error when applying the SHACL shapes, but should not

It's particularly difficult to take into account reasoning as you might want to state that certain generic relations/classes should not be asserted as they can be inferred. To be honest, I would discourage the unlimited creation of new subrelations as relations define the query patterns and not every RDF engine supports reasoning. You'll notice that I've actually tried to orient as much shapes as possible to check upon the relations, as they define the graph (and thus query) patterns. Classes are implicitly there, but would require reasoning as an additional step that can be avoided in many cases. 

BTW, you say you've added `my:hasExactGeometry` and `my:hasPointGeometry` in your own ontology? Based on the naming, this would imply that you have different geometry serializations that are not 1-1 identical? Didn't we agreed that multiple serializations for one geometry node is only allowed if the CRS and coordinates are exactly the same for all serializations? That's actually what the warning is about in `<S1-b-hasGeometry-hasSerialization-sub>`

> Shape S32-geometryClass-hasSerialization-sub forces only one geometry serialization for a Geometry and punishes this with a violation error, but S1-b-hasGeometry-hasSerialization-sub just gives a warning. Should we maybe change S32-geometryClass-hasSerialization-sub to check for at least one outgoing geo:hasSerialization?```

`<S1-a-hasGeometry-hasSerialization-sub>` would trigger a violation error (`minCount 1`), while `<S1-b-hasGeometry-hasSerialization-sub>` would trigger a warning (`maxCount 1`). In my opinion the same approach should be followed for `<S32-geometryClass-hasSerialization-sub>` (split in a violation error shape and a warning shape). The examples would need updating as well for S32.

>Tidy up SHACL rules
 Shift some SHACL rules to the new community repository

[SHACL rules](https://www.w3.org/TR/shacl-af/#rules-execution) = inferencing (expanding the graph) vs SHACL constraints = validating => are you sure you meant `SHACL rules`?

situx:
Hi @mathib and welcome back!

First of all, I should clarify that this dataset has been used for the GeoSPARQL 1.0 benchmark and I wanted to see how it validates against the SHACL rules in preparation for this dataset to be adjusted to be used in a GeoSPARQL 1.1 benchmark.

Concerning the new derived properties we defined: They could be replaced by geo:hasGeometry and geo:hasCentroid relations for a GeoSPARQL 1.1 compatible dataset. However, when we do that, we still have two geometry representations for the same feature, one assigned with geo:hasGeometry and one with geo:hasCentroid. But that should be valid according to the standard.

And yes, in our GeoSPARQL 1.0 benchmark dataset we have assigned each Geometry a GML and a WKT literal of equivalent content, hence the warning which is totally fine. What I did not expect was the violation because of the newly defined properties which led me to question how users might deal with such a situation. Should they apply reasoning beforehand or should they not use other derived properties at all? The answer to this I see in some guidance we give to the users in the to-be-defined SHACL annex which is in a draft pull request right now and to which you are very welcome to contribute.

Concerning S32-geometryClass-hasSerialization-sub I agree with you that we should split it up.

Concerning TODOS, not I did not mean rules, that was a mishap and is now corrected.

mathib:
> What I did not expect was the violation because of the newly defined properties which led me to question how users might deal with such a situation. Should they apply reasoning beforehand or should they not use other derived properties at all?

Honestly, I think we should limit the number of new relations (object properties) that people can create, when extending GeoSPARQL. If we don't, datasets in the wild might end up with an unlimited amount of new relations which can seriously complicate querying, certainly in the case of data federation (reasoning might help, but it's always a double edged sword: some (parts of) queries become easier, while others might become harder). I think we should make a stance to avoid the extension of object properties, as they will declare the querying patterns. The SHACL would assure that certain patterns are there, paving the way to reliable querying.

> guidance we give to the users in the to-be-defined SHACL annex

Totally agree that this will be useful!

mperry455:
I am in favor of removing any shapes that generate a violation when reasoning is applied.
I also have other issues with S21 and S32. These shapes seem to prevent Features and Geometries from existing by themselves. That is, every Feature should have at least one Geometry, and every Geometry should be associated with a Feature. I feel like this is too restrictive and it should be perfectly fine to have unconnected Features or Geometries.

oldskeptic:
@mperry455
> These shapes seem to prevent Features and Geometries from existing by themselves. That is, every Feature should have at least one Geometry, and every Geometry should be associated with a Feature. I feel like this is too restrictive and it should be perfectly fine to have unconnected Features or Geometries.

Agree with Matt, there are a number of use cases where a Feature is known to exist, ex: a city or a store, but its geometry may not be immediately available.

dr-shorthair:
Yes - we have a separate geometry-store. I guess you could assert there should be no geometry that isn't associated with a feature, but there may also be some utility in creating some systematic geometry sets ahead of them being attached to features - e.g. DGGS?

mperry455:
Based on this discussion and yesterday's meeting, I created #244 to remove shapes that prevent standalone Geometries and Features.

nicholascar:
I agree to these changes: if a particular system wants to ensure all Geometries are associated with Features, they can create specific Shapes for this but the ontology should not require it.

nicholascar (approved):

mperry455 (approved):

PR #227, opened 2021-10-11
Changed class definitions from rdf:Class to rdfs:Class

situx:
Just some bugfixes, as there is no rdf:Class, just an rdfs:Class to my knowledge


nicholascar:
Good spotting!

PR #235, opened 2021-11-17
Issue 228

### LINKED ISSUE #228: Example B 1.2.2 missing Z coordinate
I have the following questions/problems with example B1.2.2:

1. In example B1.2.2, the geometry is missing a Z coordinate and is not defined as POLYGON Z in the WKT literal.
It is not necessarily a problem per se, but I find it confusing that the feature defines a volume but does not provide a 3D geometry. This case can occur, but I wonder if it should be illustrated in this example.
2. In addition, it poses the question: What do we do if we have a 3D and a 2D representation of the same geometry?
Would it make sense to define a relation geo:has3DGeometry?
3. Also if two geometries are related to the geo:Feature, e.g. geo:hasDefaultGeometry relates to a 2D representation and geo:hasGeometry relates to a 3D representation, which value do we assign to a geo:hasBoundingBox relation attached to the feature representation? The 2D or the 3D bounding box?
4. The same question for the geo:hasCentroid property
5. Suppose we assign two centroids and two boundingboxes to the feature. Do we need to relate them to the respective geometry somehow?

Clearly, there is also a workaround for some of the points 3-5: Many properties like geo:hasVolume can also be used at a Geometry instance. However, the geo:hasBoundingBox and geo:hasCentroid properties are for now defined in the domain of geo:Feature, so for those the points raised in 3.-5. are still valid.
Do we leave this up to the users to decide?

nicholascar:
Partly addresses Issue #228

Closes #228

### DIFF
--- a/1.1/spec/18-Annex-C.adoc
+++ b/1.1/spec/18-Annex-C.adoc
@@ -390,6 +390,8 @@ eg:x
 
 The properties defined for this example's `Feature` instance are vaguely aligned in that the values are not real but are not unrealistic either. It is outside the scope of GeoSPARQL to validate `Feature` instances' property values.
 
+Note that this `Feature` has a 2D `Geometry` and yet a property indicating a scalar volume: `geo:hasVolume`. Used in this way, the scalar property is indicating information that cannot be calculated from other information about the `Feature` such as its geometry. Perhaps a volume for the feature has been esimated or measured in such a way that a 3D geometry was not created.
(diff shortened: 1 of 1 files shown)


situx:
This is fine with me, as we have agreed to postpone further changes to GeoSPARQL 1.2

situx (approved):

mperry455 (approved):

Issue #201, opened 2021-08-28
Clarify computations on geodetic coordinates


FransKnibbe:
Some Coordinate References Systems are geodetic; they use degrees of latitude and longitude to indicate positions on a sphere-like object (such as Earth). GeoSPARQL is unclear on how computations involving geodetic coordinates should be performed. Take the function `geof:distance` for example. It returns the shortest distance between geometries, using the CRS of the geometry in the first argument. It would be logical to assume that a SPARQL engine supporting GeoSPARQL looks up the CRS IRI to determine if the CRS is geodetic, and use a haversine distance computation if it is. And computing cartesian distance would be wrong. But I would not be surprised if GeoSPARQL implementations always compute cartesian distances.

I think the specification should be more clear on geodetic vs. cartesian CRSs, and be clearer about expected output from functions that can work on geodetic or cartesian CRSs.

jabhay:
Ensure this is covered off by #212

PR #245, opened 2021-12-06
Mention role of CRS in geometric computations

FransKnibbe:
Adds a paragraph about the role of CRS in geometric computations to the Geometry extension section.
Resolves [issue 201](https://github.com/opengeospatial/ogc-geosparql/issues/201).


FransKnibbe:
@situx: I copied the hyperlink from elsewhere in the specification. I  see now that the actual hyperlink (http://www.opengis.net/def/crs/OGC/1.3/CRS84) is correct, but its text label (http://www.opengis.net/def/crs/OGC/1.3/84) is not. The part "CRS" is missing before "84". Both occurrences (I could find no other) are fixed now.

nicholascar (approved):

PR #250, opened 2021-12-14
license & rights added to all assets

### LINKED ISSUE #60: Missing license
This issue [was reported in the OGC standards tracker](http://ogc.standardstracker.org/show_request.cgi?id=668). 

When the current GeoSPARQL ontology is [run through DBpedia Archivo](https://archivo.dbpedia.org/info?o=http://www.opengis.net/ont/geosparql), two missing things arise: a lack of a license and non-conformity with LODE. This issue about the missing license (see [issue 63](63) about LODE-conformance).

In general, the purpose of a license is to clarify what people are allowed to do with an ontology. A missing license could lead to either misuse of the ontology, or hesitance in putting the ontology to use.
See also [these remarks on licensing by the Spatial Data on the Web Working Group](https://www.w3.org/2015/spatial/wiki/Ontology_rights_and_licence).

I would like to ask for first discovering an appropriate license for the GeoSPARQL ontology and then adding it to the ontology.

nicholascar:
Closes #60

Although we haven't attained a full RDF license, we have a stable reference in all assets for now. We will replace this with an RDF reference IF the OGC presents one before 1.1 is released, if not, the RDF version can be referenced in future GeoSPARQL work.

### DIFF
--- a/1.1/funcsrules.ttl
+++ b/1.1/funcsrules.ttl
@@ -19,6 +19,8 @@
     owl:imports <http://www.opengis.net/def/ogc-na> ;
     owl:versionIRI <http://defs-dev.opengis.net/static/definitions/conceptschemes/functions_geosparql.ttl> ;
     skos:definition "A vocabulary (taxonomy) of the functions and rules defined within the GeoSPARQL 1.1 specification"@en ;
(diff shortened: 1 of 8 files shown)


situx (approved):

jabhay (approved):
Thanks for adding this Nick

PR #252, opened 2021-12-14
Geom collection

### LINKED ISSUE #203: GeometryCollections of GeometryCollections?
[PR 150](https://github.com/opengeospatial/ogc-geosparql/pull/150) has added the class `geo:GeometryCollection` to the ontology. However, the [Simple Features specification](https://portal.ogc.org/files/?artifact_id=25355), on which GeoSPARQL is based, already contains a class GeometryCollection. That means GeoSPARQL implementations would have to deal with two slightly different collections of Geometries. And it means we could have RDF GeometryCollections of SFA GeometryCollections.
Isn't that very complicated?  What is the reason `geo:GeometryCollection` exists if there already is a way to have collections of Geometries?

nicholascar:
I've added the clarification note requested in #203.

Closes #203


situx (approved):

1.1/spec/11-Part-08.adoc
```diff
@@ -71,6 +71,15 @@ geo:GeometryCollection
 
 The restriction imposed on the more general <<Class: SpatialObjectCollection, Spatial Object Collection>> that defines this class is that only instances of <<Class: Geometry, Geometry>> are allowed to be members of it and these are to be indicated with the `rdfs:member` property.
 
+[NOTE]
+====
+Some geometry literal formats, such as WKT, GML and GeoJSON, allow for collections of geometries. In WKT's case, this is indicated with the geometry type _GeometryCollection_. While the names of these objects are the same and this class' and the concepts are similar, there is no RDF/ontology relationship between literals of this type and this class. Since this class contains whole `geo:Geometry` instances, they may have much more information than geometry serialization within a literal geometry collection.
```
FransKnibbe, line 76: I think it is good to mention the other serialisations here as too: GML and KML have MultiGeometry, GeoJSON and WKT have GeomertryCollection.

1.1/spec/11-Part-08.adoc
```diff
@@ -71,6 +71,15 @@ geo:GeometryCollection
 
 The restriction imposed on the more general <<Class: SpatialObjectCollection, Spatial Object Collection>> that defines this class is that only instances of <<Class: Geometry, Geometry>> are allowed to be members of it and these are to be indicated with the `rdfs:member` property.
 
+[NOTE]
+====
+Some geometry literal formats, such as WKT, GML and GeoJSON, allow for collections of geometries. In WKT's case, this is indicated with the geometry type _GeometryCollection_. While the names of these objects are the same and this class' and the concepts are similar, there is no RDF/ontology relationship between literals of this type and this class. Since this class contains whole `geo:Geometry` instances, they may have much more information than geometry serialization within a literal geometry collection. 
+
+GeoSPARQL encourages the use of the class `geo:GeometryCollection` in situations where this class and a geometry collection literal seem to be equivalent.
```
FransKnibbe, line 78: How can this recommendation be explained? GeoSPARQL is based on the SF specification, which defines GeometryCollection. So sf:GeometryCollection already is an integral part of GeoSPARQL. Also, spatial data engines and GeoSPARQL functions are based on SF. That means that sf:GeometryCollections can act as input or output of GeoSPARQL functions. An RDF collection can not do that. I feel that rather than recommending one or the other we should explain how each type of collection can be used.

1.1/spec/11-Part-08.adoc
FransKnibbe: It seems to me that a main reason for collections of geometries existing in geometry serialisations is that it is a feature of  the SF specification. So maybe start the note with that?

1.1/spec/11-Part-08.adoc
```diff
@@ -71,6 +71,15 @@ geo:GeometryCollection
...
+
+There is also no RDF/ontology relationship between this `geo:GeometryCollection` class and the Simple Features Vocabulary's `sf:GeometryCollection` class since the former is a collection of `geo:Geoemtry` objects and the latter is to be used for compound geometry literals, as per WKT's_GeometryCollection_.
```
FransKnibbe, line 80: typo: geo:Geoemtry

mperry455 (approved):

1.1/spec/11-Part-08.adoc
nicholascar: Thanks for point out these issues. Excellent eye for detail @FransKnibbe! I have updated the NOTE to reflect your suggestions.

1.1/spec/11-Part-08.adoc
nicholascar: As above comment.

1.1/spec/11-Part-08.adoc
nicholascar: fixed

nicholascar:
@FransKnibbe please re-review as I think I've addressed your concerns.

FransKnibbe (approved):
Thanks. This should do the trick.

PR #265, opened 2022-01-04
Update definitions of collections

FransKnibbe:
Bring collection definitions inline with conventions; no need to repeat type or label in the definition.


situx (approved):

nicholascar (approved):
Thanks again @FransKnibbe

Issue #267, opened 2022-01-05
geo:ehEquals equals geo:rcc8eq


FransKnibbe:
Other than the name (IRI) there seems to be no difference between properties `ehEquals` and `rcc8eq`. Shall we make this official by using `owl:sameAs` to link the two?

dr-shorthair:
Else `owl:equivalentProperty` - more explicit

FransKnibbe:
While translating to Dutch (working from top to bottom) I just came across the simple features topological relationships. It has an equals relation too, And possible other properties can be declared equivalent as well. I will update [PR 268](https://github.com/opengeospatial/ogc-geosparql/pull/268).

PR #268, opened 2022-01-06
declare equivalence of topological properties

FransKnibbe:
Declare equivalence of topological relationships in different families. Resolves [issue 267](https://github.com/opengeospatial/ogc-geosparql/issues/267).

### DIFF
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -252,7 +252,7 @@
 
 :ehEquals
 	a rdf:Property, owl:ObjectProperty ;
-	owl:equivalentProperty :rcc8eq, :sfEquals ;
+	owl:equivalentProperty :sfEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy
@@ -467,7 +467,6 @@
 
 :rcc8eq
 	a rdf:Property, owl:ObjectProperty ;
-	owl:equivalentProperty :ehEquals, :sfEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
@@ -586,7 +585,7 @@
 
 :sfEquals
 	a rdf:Property, owl:ObjectProperty ;
-	owl:equivalentProperty :ehEquals, :rcc8eq ;
+	owl:equivalentProperty :ehEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -238,6 +238,7 @@
 
 :ehDisjoint
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :sfDisjoint ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy
@@ -251,7 +252,7 @@
 
 :ehEquals
 	a rdf:Property, owl:ObjectProperty ;
-	owl:equivalentProperty :rcc8eq ;
+	owl:equivalentProperty :rcc8eq, :sfEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy
@@ -278,6 +279,7 @@
 
 :ehMeet
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :sfTouches ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy
@@ -465,7 +467,7 @@
 
 :rcc8eq
 	a rdf:Property, owl:ObjectProperty ;
-	owl:equivalentProperty :ehEquals ;
+	owl:equivalentProperty :ehEquals, :sfEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
@@ -570,6 +572,7 @@
 
 :sfDisjoint
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :ehDisjoint ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
@@ -583,6 +586,7 @@
 
 :sfEquals
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :ehEquals, :rcc8eq ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
@@ -622,6 +626,7 @@
 
 :sfTouches
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :ehMeet ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 
--- a/1.1/geo.ttl
+++ b/1.1/geo.ttl
@@ -251,6 +251,7 @@
 
 :ehEquals
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :rcc8eq ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy
@@ -464,6 +465,7 @@
 
 :rcc8eq
 	a rdf:Property, owl:ObjectProperty ;
+	owl:equivalentProperty :ehEquals ;
 	rdfs:domain :SpatialObject ;
 	rdfs:range :SpatialObject ;
 	rdfs:isDefinedBy 


situx:
I have some improvement suggestions and partially disagree with this change:

If you look at the GeoSPARQL specification section 7 and following then rcc8eq and ehEquals apply to different geometry types, so it would appear rcc8eq is only applicable to 2D geometries whereas ehEquals is also applicable to 1D geometries.
Is that not a problem if we want to set these properties equivalent?

If it is not specified one could think about relating the SF and eh properties though, i.e.
geo:sfEquals owl:equivalentProperty geo:ehEquals
geo:sfEquals owl:equivalentProperty geo:ehEquals

nicholascar:
We do have to be careful here regarding too much ontology! 

Unless there's really a possibility that these changes will be useful, we should not add the equivalencies due to the risks @situx mentions above. 

My expectation is that there's probably not much advantage to be gained here as people surely tend to use "families" of properties, e.g. SF, or EH, or RCC8 but that they don't mix them much. Do you think this assumption's correct?

FransKnibbe:
> If you look at the GeoSPARQL specification section 7 and following then rcc8eq and ehEquals apply to different geometry types, so it would appear rcc8eq is only applicable to 2D geometries whereas ehEquals is also applicable to 1D geometries.
> Is that not a problem if we want to set these properties equivalent?

Yes, that would be a problem. But what exactly is the difference between rcc8eq  and ehEquals? What does A/A mean in table 3 of section 7? And where can information on 2D versus 1D geometries be found?

I have just noticed table 4 in section 7 for the first time. It declares rcc8eq and ehEquals to be equivalent. And there some equivalent properties that I have missed. Depending on how this discussion develops I can add them.

situx:
A/A is explained at the beginning of section 7:
"In the following descriptions, the notation X/Y is used denote applying a spatial relation to geometry types X and Y"
and further:
"the symbol A is used for 2-dimensional geometries (e.g. polygons)"
So if rcc8eq is assigned with A/A it only applies to 2D geometries and e.g. not to lines
ehEquals is assigned to "All" so it should apply to all geometry types

FransKnibbe:
> We do have to be careful here regarding too much ontology!
> 
> Unless there's really a possibility that these changes will be useful, we should not add the equivalencies due to the risks @situx mentions above.
> 
> My expectation is that there's probably not much advantage to be gained here as people surely tend to use "families" of properties, e.g. SF, or EH, or RCC8 but that they don't mix them much. Do you think this assumption's correct?

Of course we need to find out first which properties really are 100% equivalent.
But I do think expressing equivalence in the ontology is useful. I can think of two reasons:

1. Helping data publishers looking for expressions of topological relations. When looking for an expression for spatial equality, for instance, one can find three different properties. It helps to show that they are basically the same.
2. Helping data consumers find data. If the properties are used in SPARQL queries, a reasoner can make use of explicit equivalence. Or if without a reasoner, data consumers could look for equivalent combinations.

FransKnibbe:
> A/A is explained at the beginning of section 7:
> "In the following descriptions, the notation X/Y is used denote applying a spatial relation to geometry types X and Y"
> and further:
> "the symbol A is used for 2-dimensional geometries (e.g. polygons)"
> So if rcc8eq is assigned with A/A it only applies to 2D geometries and e.g. not to lines
> ehEquals is assigned to "All" so it should apply to all geometry types

Got it, thanks. That means all RCC8 properties can not take part in declarations of equivalence. I will make the required changes.

By the way, do we have something in the documentation about application of topological relationships to 3D spatial objects?
Another by the way: The new B.1 Functions Summary Table list RCC8 _functions_ as applicable to 3D Geometries. Are we certain that is correct?

situx:
I do think there is an issue still open about examples in 3D about that.

As for the RCC concerns I will cite the original paper: https://link.springer.com/content/pdf/10.1023/A:1009712514511.pdf in section 3:
"We should make precise exactly what counts as a region. In our intended interpretation
the regions may be of arbitrary dimension, but they must all be the same dimension and
must not be of mixed dimension (for example, a region with a lower dimensional spike
missing or sticking out is not intended). Such regions are termed regular. Normally, of
course our intended interpretation will be 3D, though in many of the figures in this paper,
for ease of drawing, we will assume a 2D world (as is also usual in GIS applications)."

That means even though the RCC has only been defined in 2D contexts in the old GeoSPARQL because the old GeoSPARQL dealt with 2D only, the RCC itself can be applied to 3D or even nD just not to 1D

FransKnibbe:
> As for the RCC concerns I will cite the original paper: https://link.springer.com/content/pdf/10.1023/A:1009712514511.pdf in section 3:
> "We should make precise exactly what counts as a region. In our intended interpretation
> the regions may be of arbitrary dimension, but they must all be the same dimension and
> must not be of mixed dimension (for example, a region with a lower dimensional spike
> missing or sticking out is not intended). Such regions are termed regular. Normally, of
> course our intended interpretation will be 3D, though in many of the figures in this paper,
> for ease of drawing, we will assume a 2D world (as is also usual in GIS applications)."

Thank you, that is very useful to know. So of the three topology families in GeoSPARQL, Egenhofer (eh) and Simple Features (sf) can be applied to 1 or 2 dimensional Spatial Objects, and RCC8 (rcc) can be applied to 2 or 3 dimensional Spatial Objects. I think that would be useful to put in the documentation.

>We should probably add a SHACL shape to check whether the dimensions of Features connected with rcc* properties are the same.

You mean Geometries? As far as I can tell, we have no way of knowing or specifying the dimensionality of a Feature. And we have to take into account that a Feature could have both 2D and 3D Geometries, or no Geometry at all.

situx:
Sorry, of course, I mean Geometry, you are correct.

As for the other remarks:
RCC8: 2D and higher dimensions
EH Relations 0D-2D
SF Relations: The Simple Feature Access standard states: "The instantiable subclasses of Geometry defined in this Standard are restricted to 0, 1 and 2-dimensional geometric objects that exist in 2, 3 or 4-dimensional coordinate space (ℜ2, ℜ3 or ℜ4)"
That means sf relations can also be between 3D objects

FransKnibbe:
Another thing: If the Egenhofer and Simple Features families of topological relationships can not be applied to 3D Geometry, isn't funny that comparable functions _are_ applicable to 3D Geometry, according to the [Functions Summary Table](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_functions_summary_table)?

situx:
Correct, we should remove 3D compatibility of the EH relation functions

FransKnibbe:
> SF Relations: The Simple Feature Access standard states: "The instantiable subclasses of Geometry defined in this Standard are restricted to 0, 1 and 2-dimensional geometric objects that exist in 2, 3 or 4-dimensional coordinate space (ℜ2, ℜ3 or ℜ4)"
> That means sf relations can also be between 3D objects

It important to be clear about the distinction between the dimensionality of a Geometry and the dimensionality of its coordinate space. I would call a point with coordinates (x,y,z) a 0-dimensional Geometry in 3-dimensional space.
It seems to me SF topology does not go above 2 dimensional Spatial Objects.

situx:
If this is really the case, then we would need new properties to express relations between 3D geometries or only permit rcc8 relations between them

FransKnibbe:
I have now removed declarations of equivalence involving RCC8 (because applicable dimensionality does not match the Egenhofer and Simple Features families).

FransKnibbe:
> Correct, we should remove 3D compatibility of the EH relation functions

In [the same table](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_functions_summary_table) I see other differences between the topological properties and related functions. For example, `geof:sfContains`  has valid input types "1x Polygon, 1x Geometry". Why is that?

situx:
> > Correct, we should remove 3D compatibility of the EH relation functions
> 
> In [the same table](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_functions_summary_table) I see other differences between the topological properties and related functions. For example, `geof:sfContains` has valid input types "1x Polygon, 1x Geometry". Why is that?

Because we expect the second argument to be contained by sth. that is supposed to be an area, i.e. can contain sth.
I am not sure if e.g. a point in a LineString counts as contained. If so, then we would need to add a LineString. And is a point at the same coordinates of another point contained by it? I am not sure

FransKnibbe:
> > In [the same table](https://opengeospatial.github.io/ogc-geosparql/geosparql11/spec.html#_b_1_functions_summary_table) I see other differences between the topological properties and related functions. For example, `geof:sfContains` has valid input types "1x Polygon, 1x Geometry". Why is that?
> 
> Because we expect the second argument to be contained by sth. that is supposed to be an area, i.e. can contain sth. I am not sure if e.g. a point in a LineString counts as contained. If so, then we would need to add a LineString. And is a point at the same coordinates of another point contained by it? I am not sure

The [PostGIS documentation](https://postgis.net/workshops/postgis-intro/de9im.html) has a nice explanation on how linestrings and points act in DE-9IM. But I think this line of thought does not need to be followed. Perhaps a point can never contain a polygon, but does that mean the combination of input parameters is invalid? You _will_ be able to get a valid boolean output. 

It seems to me that the only input restrictions for topology functions should be dimensionality of input Geometry (0-2 for Egenhofer and SF, 2-3 for RCC8). Otherwise all Geometry types are allowed.

FransKnibbe:
> If this is really the case, then we would need new properties to express relations between 3D geometries or only permit rcc8 relations between them

That will be difficult to do without Geometry types in the ontology. I hope they will be inlcuded in a next version (SF types and 3D types). For now, wouldn't it suffice to clearly document limitations on dimensionality for both the topology properties and topology functions? It could be done in both the specification document and the ontology (using skos:scopeNote for example).

situx (approved):

nicholascar (approved):

PR #279, opened 2022-01-23
Fixed translations to conform to the more general definition of Feature

### LINKED ISSUE #264: translating ¨Feature¨ in Dutch and German
One of the hardest things to translate in GeoSPARQL is the word ¨Feature¨. It is the type of word you see in models where there is a need for a very broad category, similar to words like _thing_, _object_ or _entity_. I guess the term _feature_ was picked by the OGC because in a certain sense of the word it implies geography, see [this Cambridge dictionary page](https://dictionary.cambridge.org/dictionary/english/feature). Probably the three meanings in English (a part of a face; an aspect of something; a geographic occurence) are related, in all cases it is about something that catches the eye.

The fact that geography is implied in the word _feature_ made it easy to use a broad definition of the class `Feature` in the GeoSPARQL ontology. Not only geographical things can be considered Features, but distinct spatial things too. A spoon or a tennis ball for example. However, the current German translation makes geography explicit:
`skos:prefLabel "Geografisches Merkmal"@de`
Can the German word _Merkmal_ imply geography, like one of the meanings of _feature_ in English? Or does it only mean a characteristic of something? Could _Geografisches_ be removed?

In Dutch I am struggling with finding a good translation. I am contemplating _voorkomen_ and _verschijnsel_, with a preference for the former, because it feels easier to accomodate both natural and artificial occurences.

Chances are that a similar word can be found in both Dutch and German.

situx:
Closes #264

### DIFF
--- a/1.1/translations/geo_de.ttl
+++ b/1.1/translations/geo_de.ttl
@@ -267,7 +267,7 @@
 :FeatureCollection
 	skos:prefLabel "Merkmalssammlung"@de ;
 	skos:altLabel "Feature Collection"@de ;
-	skos:definition "Die Klasse geo:FeatureCollection beschreibt jegliche Sammlungen von einzelnen geographischen Merkmalen."@de .
+	skos:definition "Die Klasse geo:FeatureCollection beschreibt jegliche Sammlungen von einzelnen räumlichen Merkmalen."@de .
 	
 :Geometry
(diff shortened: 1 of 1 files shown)


FransKnibbe:
Approved!

FransKnibbe (approved):

lvdbrink:
Approved!

lvdbrink (approved):

PR #287, opened 2022-02-20
Update validator.ttl

### LINKED ISSUE #177: Decide on SHACL shape about serializations
This issue is to discuss SHACL validation shape number 1.

I encourage everyone to add pro/con arguments to add to this post so that we can make an informed decision.
Please edit this post whenever needed:

1 serialization per Geometry instance:
- The coordinate reference system is the same
- There are no conflicts concerning a differing accuracy a.s.o. because only one serialization per geometry exists

Many serializations per Geometry:
- The serializations need to be equivalent, but not equal. For example, they could be in different CRS according to literal restrictions (GeoJSON only allows WGS84 for example)
-  The serializations could have a differing accuracy

Please add more arguments for either side that I missed

FransKnibbe:
Remove shape 1a; make shapes 1 and 3 work on solo Geometries too. Solves[ issue 177](https://github.com/opengeospatial/ogc-geosparql/issues/177).

Fixes #177

### DIFF
--- a/1.1/validator.ttl
+++ b/1.1/validator.ttl
@@ -30,83 +30,38 @@ As of GeoSPARQL 1.1, this validator is not normative, only informative, however 
 .
 
 #################################################
-# Shape 1 (1a-1e)
+# Shape 1 (1a-1d)
 #################################################
+
+<S1-Geometry>
+	a sh:NodeShape ;
+	sh:targetClass geo:Geometry ;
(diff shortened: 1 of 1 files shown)


situx (changes requested):
I think your changes work in general, but I think we would also need to update the table in the GeoSPARQL Annex. Was that part of your pull request or should we create a different pull request for that?

jabhay:
- [ ] Update table in Appendix E to reflect changes to shapes.

nicholascar (approved):
I'm approving and merging but will submit a follow-up PR to apply some further fixes

FransKnibbe:
> I'm approving and merging but will submit a follow-up PR to apply some further fixes

Hi @nicholascar: Will that follow-up include an update of Annex D.3?

PR #293, opened 2022-03-08
fix Table B1 formatting

nicholascar:


nicholascar:
Trivial PR: formatting only, so I'm merging without review.

PR #305, opened 2022-03-15
SHACL test suite

nicholascar:
I have fund a few bugs with our SHACL files and I've even found a bug with the pySHACL tool I usually use for SAHCL validation, which I've now fixed.

To ensure we have confidence in our SHACL shapes, I've created a test suite for them which contains valid & invalid example data for each shape.

The SHACL file has change a lot but there's been no change in scope: I've only patched things I could show to be errors through the valid & invalid data files and added in some missing tests - various formats that weren't present in GeoSPARQL 1.1 when the first Shapes file version was made.

### DIFF
--- a/1.1/examples/shacl/S04-invalid-01.ttl
+++ b/1.1/examples/shacl/S04-invalid-01.ttl
@@ -1,4 +1,4 @@
-# Invalid: the geometry literal doesn;t have a datatype matching the property that indicates it
+# Invalid: the geometry literal doesn't have a datatype matching the property that indicates it
 @prefix geo: <http://www.opengis.net/ont/geosparql#> .
 
 <feature-z>
--- a/1.1/examples/shacl/S04-invalid-01.ttl
+++ b/1.1/examples/shacl/S04-invalid-01.ttl
@@ -1,0 +1,8 @@
+# Invalid: the geometry literal doesn;t have a datatype matching the property that indicates it
+@prefix geo: <http://www.opengis.net/ont/geosparql#> .
+
+<feature-z>
+    geo:hasGeometry [
+        geo:asWKT "POINT (153.084230 -27.322738)" ;
+    ] ;
+.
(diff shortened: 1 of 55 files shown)


situx:
A problem I have encountered before is that I guess pySHACL was not able to cope with multiline literals. Maybe we could add a test for that here?

nicholascar:
I think this _may_ be solved since I've added `sh:flags "m" ;` to the patterns that need to match multi-line literals. So now literals such as this pass validation:

```
        geo:asGML """<gml:Point
                       srsName=\"http://www.opengis.net/def/crs/OGC/1.3/CRS84\"
                       xmlns:gml=\"http://www.opengis.net/ont/gml\">
                   <gml:pos>-83.38 33.95</gml:pos>
               </gml:Point>"""^^geo:gmlLiteral ;
```

situx (changes requested):
S15-invalid-01 uses the geo:hasSpatialResolution property in the wrong way I believe.
This property needs a unit description according to the specification.
The same with s14-valid.
And should not S15 actually test for Spatial Accuracy instead of Spatial Resolution?
In any case, also the Spatial Accuracy property would need a unit, I think

nicholascar:
I've addressed @situx comments with a subsequent commit.

situx (approved):

1.1/examples/shacl/S04-invalid-01.ttl
```diff
@@ -0,0 +1,8 @@
+# Invalid: the geometry literal doesn;t have a datatype matching the property that indicates it
```
mperry455, line 1: typo: doesn;t -> doesn't

1.1/examples/shacl/S15-invalid-01.ttl
```diff
@@ -0,0 +1,16 @@
+# Invalid: More than one hasSpatialResolution property
```
mperry455, line 1: typo: hasSpatialResolution -> hasSpatialAccuracy

mperry455 (approved):

nicholascar:
Thanks @mperry455 for the review and fixes! I'm merging in now after these two approvals

PR #318, opened 2022-04-14
Update E.13 Ordnance Survey mapping

PeterParslow:
We are pleased to have contributed to the discussion some ten years ago but recognise that the subject area has moved on. We would not recommend people starting to relate to our ontology now, and we look forward to migrating to some more authoritative one in due course.

Also, eastings & northings are best not confused with latitudes & longitudes. I have put in the definitions from the OS ontology: they are metres from an origin, in a defined "flat & square" projection.

I'm glad you've ignored the grid square classes in our ontologies!

### DIFF
--- a/1.1/spec/20-Annex-E.adoc
+++ b/1.1/spec/20-Annex-E.adoc
@@ -328,7 +328,8 @@ https://github.com/openplannerteam/routable-tiles-ontology
 |===
 
 == E.13 Ordnance Survey UK Spatial Ontology
-
+Note: these two ontologies will be withdrawn during 2022.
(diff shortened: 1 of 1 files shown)


situx:
Hi Peter, 
thanks for the PR and corrections from your perspective.
I agree we should put a deprecated soon note in as you have been suggesting.
Does that mean you want to migrate to GeoSPARQL?

situx (approved):

nicholascar (approved):

nicholascar:
Thanks @PeterParslow for the update. I'll merge in now and you/OS have plenty of time to make any further updates still as we are not yet even in public review for this GeoSPARQL version yet.

PR #320, opened 2022-04-19
re-mapping to BFO2020 due to feedback

nicholascar:
Due to BFO authors pointing to BFO 2020 as the best target


nicholascar:
image
image

situx (approved):
The mapping looks ok to me. However, it again somehow underlines to me that in GeoSPARQL 1.2, we should give some guidance as to how we expect people to deal with spatiotemporal data.

dr-shorthair:
Here's the BFO class hierarchy from ISO/IEC 21838-2:2021 for reference
image

dr-shorthair (approved):

nicholascar:
Yes, I think that GeoSPARQL's lack of temporal handing is really shown up here! We do need to write about it in 1.1 and then to take it very seriously for 1.2.

mperry455 (approved):

Issue #272, opened 2022-01-18
Replace SRS with CRS, where appropriate


FransKnibbe:
The GeoSPARQL document seems to regard Spatial Reference System (SRS) as a synonym of Coordinate Reference System (CRS). For example, here is the definition of `geof:buffer`:

> Returns a geometric object that represents all Points whose distance from geom1 is less than or equal to the radius measured in units. Calculations are in the spatial reference system of geom1.

I think it would be good to change this kind of occurrence of "spatial reference system" to "Coordinate Reference System", at least in the cases where it applies to Geometry, because that is the intended meaning.

situx:
I get your point, but I do not see it as something that needs to be changed.
My interpretation is that a spatial reference system is a broader specification than a coordinate reference system.
In the future, we might have GeoCoding Literals as input of buffer which might not have a coordinate reference system but rather a spatial reference system.
So following that definition I see no problem leaving it as spatial reference system

FransKnibbe:
Hi @situx, I have updated the issue a bit, to make clear that SRS can be changed to CRS only in cases where is applies to geometry.
You are right that SRS is broader than CRS. It is strange and misleading to use the broader term when only the narrower term applies. And it is very likely that the original authors intended to indicate CRS rather than SRS. The concepts have historically been confused, also in OGC documents.

FransKnibbe:
Is it OK to change requirement IRIs too? Or does that break backwards compatibility?
For example, in the IRI below "srs" could be changed to "crs"
http://www.opengis.net/spec/geosparql/1.0/req/geometry-extension/wkt-literal-default-srs

situx:
@FransKnibbe Requirement IRIs have never been resolvable but have been defined in the original spec document. So in my opinion we cannot change them without breaking backward compatibility.

PR #323, opened 2022-04-20
Replace srs with crs

FransKnibbe:
This PR should replace all occurrences of "SRS" and "spatial reference system" in the specification document with "CRS" and "coordinate reference system", where appropriate. Fixes [issue 272](https://github.com/opengeospatial/ogc-geosparql/issues/272).

### DIFF
--- a/1.1/spec/01-Preamble.adoc
+++ b/1.1/spec/01-Preamble.adoc
@@ -75,9 +75,11 @@ to
 |===
 
 === Major changes between versions 1.0 and 1.1
-Version 1.1 of GeoSPARQL was released approximately 9 years after version 1.0. It contains no breaking changes to 1.0, but does contain additions: whole new profile resources, new ontology elements and new functions. The major changes are given in the tables below.
+Version 1.1 of GeoSPARQL was released approximately 9 years after version 1.0. It contains no breaking changes to 1.0, but does contain additions annd clarifications. Profile resources, ontology elements and functions were added. The major changes are given in the tables below.
 
-These new profile resources are resources - documents - that are separate from this specification. The new _profile defintion_ lists all the GeoSPARQL 1.1 resources.
+==== Profile resources
+
+These new profile resources are resources - documents - that are separate from this specification. The new _profile definition_ lists all the GeoSPARQL 1.1 resources.
 
 |===
 |New resource | Location
@@ -86,6 +88,8 @@ These new profile resources are resources - documents - that are separate from t
 |GeoSPARQL Rules in RIF | http://www.opengis.net/def/geosparql-rifrules
 |RDF validation file | http://www.opengis.net/def/geosparql-shapes
 |===
+
+==== Ontology elements and functions
 
 These new ontology elements and new functions are normatively defined in this specification document.
 
@@ -149,6 +153,12 @@ These new ontology elements and new functions are normatively defined in this sp
 |aggConcaveHull | <<Function: geof:aggConcaveHull>>
 |aggUnion | <<Function: geof:aggUnion>>
 |===
+
+==== Clarifications
+* The terms Spatial Reference System (SRS) and Coordinate Reference System (CRS) are no longer interchangeable. Spatial Reference System is now taken to be a broader category than Coordinate Reference System, see the <<Terms and definitions>>.
+* Class definitions were updated to be more self-contained and easier to understand for people without a background in geoinformatics.
+* A section was added on <<Recommendation for specification of units of measurement,the specification of units of measurement>>.
+* A section was added on the <<Influence of Coordinate Reference Systems on geometric computations>>.
 
 == v. Changes to the OGC® Abstract Specification
 The OGC® Abstract Specification does not require changes to accommodate this OGC® standard.
--- a/1.1/spec/09-Part-06.adoc
+++ b/1.1/spec/09-Part-06.adoc
@@ -418,7 +418,7 @@ SELECT (COUNT(*) AS ?cnt)
 WHERE { :f1 geo:sfDisjoint :f1 }
 ```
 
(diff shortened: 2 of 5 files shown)


situx:
I wonder if this is necessary. 
I believe leaving the term SRS in the specification except for functions or places where we explicitly require a CRS is fine.
However, I would even argue that no function prohibits the use of a non-CRS system as of now.

For example, looking at the GeoSPARQL 1.0 specification, requirements 10,11,12 (WKTLiteral), the terms are used as follows:
* Requirement 10 status: "All RDFS Literals of type geo:wktLiteral shall consist of an optional URI identifying the **coordinate reference system** followed by Simple Features Well Known Text (WKT) describing a geometric value"
All other requirements concerning WKT speak of spatial reference system URIs.

You might call this an inconsistency in GeoSPARQL 1.0, but my interpretation would be that the terms coordinate reference system and spatial reference system have been used interchangeably. Maybe @mperry455 knows more.

We should be careful about removing the scope of original requirements texts by changing this wording.

My case would be as follows: We should stick to the term spatial reference system, especially in the case when we only describe a URI. How to apply a spatial reference system that is not a coordinate reference system to an e.g. WKT Literal? That is probably an implementation issue, but as for now, I think no issue the GeoSPARQL spec has to be concerned with.
All we should say is how to identify a (spatial) reference system of whatever kind in my opinion and that is by a URI.

jabhay:
Wondering whether this could be considered a breaking change based on someone implementing an SRS that isn't a CRS, where documentation has changed from SRS to CRS. I'm happy for the group to decide this isn't the case. Keen for thoughts below :)

jabhay:
Suggestion: to add a note clarifying that the change is intended to improve the original intent, being scoped to CRS. Previously CRS and SRS were more interchangeable than they are now.

FransKnibbe:
I have just added a clarification to the Revision History (which now has a section about clarifications).

> I wonder if this is necessary. I believe leaving the term SRS in the specification except for functions or places where we explicitly require a CRS is fine. However, I would even argue that no function prohibits the use of a non-CRS system as of now.

Geometry functions can not function with a non-CRS SRS. 

> 
> For example, looking at the GeoSPARQL 1.0 specification, requirements 10,11,12 (WKTLiteral), the terms are used as follows:
> 
> * Requirement 10 states: "All RDFS Literals of type geo:wktLiteral shall consist of an optional URI identifying the **coordinate reference system** followed by Simple Features Well Known Text (WKT) describing a geometric value"
>   All other requirements concerning WKT speak of spatial reference system URIs.
> 
> You might call this an inconsistency in GeoSPARQL 1.0, but my interpretation would be that the terms coordinate reference system and spatial reference system have been used interchangeably. Maybe @mperry455 knows more.

Yes, they have been uses interchangeably. But that is no longer in line with other OGC specifications, or the updated terms & definitions in GeoSPARQL 1.1.

nicholascar:
Sorry but I cannot agree to this change as it will break DGGS work such as the implementations we've already made. 

We already changed all references to CRS _to_ SRS in previous PRs to allow for spatial referencing without CRSes, e.g. DGGS.

> Geometry functions can not function with a non-CRS SRS.

Sure, and GeoSPARQL doesn't delve into the internals of geometry literals - e.g. the elements within a GML or WKT geometry serialization - so, by the same token, if data needs CRSes to function, not SRSes, then CRSes need to be supplied by the user.

GeoSPARQL is a model that can be constrained by profiling, so if someone wants to use GeoSPARQL with CRS-only data, or with WKT only or with GML using GDA94 CRS only etc., they are free to implement profiles to enforce such. We - the WG - have even set up a community repository, https://github.com/opengeospatial/ogc-geosparql-shapes, where such constraints can be placed by the community for shared use.

> Yes, they have been uses interchangeably. But that is no longer in line with other OGC specifications, or the updated terms & definitions in GeoSPARQL 1.1.

We do not intend to use SRS & CRS interchangeably: we mean SRS and CRSes are a subset of SRSes. If there are wording issues, we should fix those.

jabhay:
I will take this up with OGC representatives to determine their view on the use of SRS to support the vote contained in #112

jabhay:
The OGC team have indicated a desire to leave the wording consistently as SRS to enable broader applicability. @ogcscotts may wish to add his thoughts.

FransKnibbe:
> The OGC team have indicated a desire to leave the wording consistently as SRS to enable broader applicability. @ogcscotts may wish to add his thoughts. As such, this PR will be rejected once the author is back on deck.

@jabhay, could you please explain the desire of the OGC team? Or better yet, quote their response here? The goal of "enabling broader applicability" is not clear to me. It it the wish of the OGC to alter the definition of Geometry?

jabhay:
@ogcscotts , could you put a response below on the following topic:

When creating the original GeoSPARQL standard, the authors used SRS rather than CRS because at the time the two terms were interchangeable. More recently, there is a clearer differentiation between them. With the OGC definition of Geometry that we are familiar with, "GM_Object instances are sets of direct positions in a particular coordinate reference system."

This is causing some contention in the group, as we have extended the use of literals to include representations of Geometry that do not have CRSs - for example DGGS. Here, the literal will include grid references rather than coordinates, and a spatial reference system rather than a co-ordinate reference system.

We are hoping for clarification on whether OGC believes using the term SRS in this context, given DGGSLiterals (and potentially other representations of Geometry) described as representing Geometries, is legitimate? Would be great to record the rationale for that.

FransKnibbe:
From the perspective of backward compatibility, this is a good PR. The definition of Geometry in GeoSPARQL 1.0 includes the aspect of geometry having a CRS. All this PR does is aligning the documentation with the definition and the definitions of SRS and CRS which have now been established.

I understand rejection of this PR to change the definition of Geometry in GeoSPARQL to no longer align with the definition of Geometry in GeoSPARQL 1,0 and in other OGC standards (for example, SFA-CA). I don't think that is a good idea and I hope people are aware of the ramifications.

Issue #310, opened 2022-04-06
How to disclose the GeoSPARQL capabilities of SPARQL endpoints


FransKnibbe:
In an automated web of data, it would be good if consumers, especially machines, have a way of establishing if and how a SPARQL endpoint support GeoSPARQL, in the form of an exchange of data.

For instance, the GeoSPARQL specification could include guidance on how to use [SPARQL Service Description](https://www.w3.org/TR/sparql11-service-description/) to make the following known:

1. Whether the endpoint supports GeoSPARQL
2. Which version(s) of GeoSPARQL are supported
3. Which GeoSPARQL modules (conformance classes) are supported

situx:
That is a very good idea.
We should definitely follow up on this.
Currently, especially in my software implementations, I use a set of test queries to find out whether GeoSPARQL is supported and to which extent, but a service description would help a lot.

situx:
I had a look at the specification vocabulary today and I will break down my insights here:

Insight number 1:
You are able to specify a query language type, which would be easy for us to do and is just one triple.
sd:supportedLanguage sd:SPARQL11Query ; is an example for SPARQL 1.1.
We could simply define two URIs for GeoSPARQL1.0 and GeoSPARQL 1.1 query language, right?
However,  I believe we currently do not have either to identify the query language itself.
We have:
* <http://www.opengis.net/def/geosparql/1.1> to identify the standard as a whole
* <http://www.opengis.net/def/geosparql/funcsrules> to identify functions and rules
So I wonder if at least a SPARQL service description with a unique URL identifying the GeoSPARQL query language is feasible.
@nicholascar do we really not have a URI to identify the query language itself as of now?


Insight number 2:
The service description allows only the specification of extension functions, not modules.
To me, it seems extension functions are only defined if they are not included in the specified supported query languages.
However, we have URIs for every query function we define in GeoSPARQL in the requirements vocabulary, so it would be no stretch to add them to a service description.

Modules are not defined in the service description vocabulary, but we already have URIs defined for modules and could assign them to the sd:feature property, as is done for example to indicate the ability for federated querying.

To sum up, we would need:
* The definition of two URIs for the GeoSPARQL 1.0 and GeoSPARQL 1.1 query languages
* A decision on whether we recommend listing of all GeoSPARQL functions in a service description
* A decision on whether it makes sense to list GeoSPARQL modules as sd:feature

[...]

PR #328, opened 2022-04-29
Added service descriptions for discussion

situx:
This pull request includes three kinds of service descriptions derived from the discussion in #310 
We should discuss if we want to include them and in which way.
Also, two URIs for the GeoSPARQL 1.0 and GeoSPARQL 1.1 query languages should be defined.


nicholascar (changes requested):
I've made a few minor RDF syntax fixes.

However, I think the return formats need to be examined a bit more closely: we have functions that return boolean values and others that return RDF objects, both of which can be packaged in either SPARQL results (JSON, CSV etc) or RDF via CONSTRUCT statements.

Also, how will these descriptions be listed in GeoSPARQL? Another Annex?

If the answers to these questions aren't very simple, I suggest deferring this addition until 1.2.

situx:
@nicholascar I think the return formats are not an issue here. I do not see them defined anywhere in a SPARQL service description. 
My proposal on how to include it is another profile entry. Whether it is worth a whole Annex, I am not sure.
The minimum we could add is two URIs for the query languages GeoSPARQL 1.0 and GeoSPARQL 1.1 and describe shortly how to use those in a SPARQL service description.
But that is of course meant for discussion in the meeting today

nicholascar:
OK, if we just go with the minimal Profile resources, we will just have to think about how to communicate their existence to others. Perhaps this is just a note somewhere in the Spec - perhaps just the Profile parts section - and then we can directly inform implementers like Jena.

jabhay:
Where in the specification should we put this?
Options:
1. Where the profile parts are listed.
2. Introduction section.
3. Function listing sections (Annex A).
4. Conformance section (Section 2).
A sentence in each of the above sections will complete the specification change.

situx:
- Add sd:Feature specifications to the requirements vocabulary
- Add two collections of conformance classes for GeoSPARQL 1.0 and 1.1 to the requirements vocabulary
- Two examples with sd:Feature and one sd:Feature with extension functions defined

1.1/funcsrules.ttl
```diff
@@ -1256,7 +1257,7 @@ rules:sfCrosses a skos:Concept ;
     skos:definition "A query rewrite rule used to determine if two spatial objects cross based their associated primary geometry objects."@en ;
     skos:prefLabel "crosses"@en .
```
lvdbrink, line 1258: Shouldn't sfDisjoint and sfTouches also be declared sd:Function?

1.1/servicedescription/servicedescription_all_functions.ttl
```diff
@@ -0,0 +1,68 @@
...
+	sd:extensionFunction geof:sfEquals;
+	sd:extensionFunction geof:sfIntersects;
+	sd:extensionFunction geof:sfOverlaps;
+	sd:extensionFunction geof:sfTouches;
+	sd:extensionFunction geof:sfWithin;
+	sd:extensionFunction geof:symDifference;
+	sd:extensionFunction geof:union;
+    sd:resultFormat <http://www.w3.org/ns/formats/RDF_XML>, <http://www.w3.org/ns/formats/Turtle> ;
+    sd:defaultEntailmentRegime ent:RDFS
+.
```
lvdbrink, line 66: As far as I can tell not all functions are listed here: metricBuffer, coordinateDimension, metricDistance, geometryType, is3D, isMeasured, spatialDimension, lenght, projectTo, and the aggregate functions are missing.

Also, minX, maxX, etc should be with a capital last letter.

lvdbrink (commented):
I think the service description would be a useful addition!

1.1/funcsrules.ttl
situx: I think the functions are, but you pointed out the RIF rule definitions. 
As far as I know there is no equivalent to a RIF rule in the Service Description vocabulary

1.1/servicedescription/servicedescription_all_functions.ttl
situx: That is correct. Something I spotted in the other pull request which closes 9 issues.
Not sure I entered them all there, but at least some of them.
So it would be better to merge that one in first

situx:
I have made the changes that were discussed in the meeting and created SPARQL 1.1 service descriptions for GeoSPARQL 1.1 and 1.0.
One question remains for me:
* May we add single conformance classes as extensions to SPARQL 1.1 service descriptions or only the modules? If we can add single conformance classes, then I would add another example file

jabhay:
Outcome from discussion in meeting: Yes, we should include the conformance classes.

1.1/servicedescription/servicedescription_all_functions.ttl
situx: projectTo was removed from the specification recently. I will add the others now

situx: Should be changed now. Thanks for checking

nicholascar (approved):
Looks good @situx

PR #326, opened 2022-04-21
Fixes and corrections collection

situx:
I will collect smaller fixes and changes that I correct in this pull request. I will link issues that will be closed by this pull request down below.
Changes:
* Added property mapping sdo:geoMidpoint owl:equivalentProperty geo:hasCentroid
* Fixed typos and minor other corrections
* Closes #322 by adding a comment on GeometryCollections to geometryN
* Closes #312  by adding a comment on the LinkedGeoData ontology
* Closes #311 by changing the function signature of geometryType to xsd:anyURI and adjusting the description
* Closes #294 by adding @en labels to ontology files
* Closes #327 by adding @en labels to SF vocabulary and by changing rdfs:comment and rdfs:label to skos:definition and skos:prefLabel
* Closes #298 by adding a note about spatiotemporal support in the introduction
* Closes #260 by removing aggConcatLines. I believe the functionality is already covered by geof:aggUnion called with only LineStrings (to be discussed)
* Closes #246 by adding the GML vocabulary to the 1.0 and 1.1 profile.ttl
* Closes #261 by adding definitions for geof:concaveHull and geof:boundingCircle

PR #329, opened 2022-05-04
move rdfs: props to skos: in SF vocab

### LINKED ISSUE #327: rdfs:comment in Simple Features vocabulary
The Simple Features vocabulary still uses rdf:comment instead of skos:definition.
Should we change that for the GeoSPARQL 1.1 release?

nicholascar:
Closes #327


nicholascar:
Duplicates part of PR #326

PR #334, opened 2022-05-18
Changed wrong GML namespace in GML example

### LINKED ISSUE #330: Correct XMLns for gml?
The GeoSPARQL standard references [OGC07-036](https://portal.ogc.org/files/?artifact_id=20509) here: https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/spec/11-Part-08.adoc#geography-markup-language-serializationgml

which defines `http://www.opengis.net/gml/3.2` as the XML namespace of GML in section 5.1

This example file from the repo:
https://github.com/opengeospatial/ogc-geosparql/blob/7ac42091a9fdf90cefe5f051f5c616fb5a1af21e/1.1/moreton-island.gml#L41-L42
(currently removed)

likewise uses this namespace.

In the sole example on
https://github.com/opengeospatial/ogc-geosparql/blob/dc9cb57d6c8bf90a116c993c0f06f3d4b99b57cb/1.1/spec/11-Part-08.adoc?plain=1#L442-L447
however, a different namespace (the RDF Ontology namespace `http://www.opengis.net/ont/gml`) is used.

This creates interoperability problems when consuming a GML file from WFS and trying to load it into a gmlLiteral, as the XML tags do not match any more. Is this a bug/bad example in the standard? Or are we supposed to massage GML with `sed` to alter the XML namespaces?

Currently, Apache Jena hardcodes the NS from the example:
https://github.com/apache/jena/blob/231624d00188f58da6c15fcf5f940f0448111d02/jena-geosparql/src/main/java/org/apache/jena/geosparql/implementation/parsers/gml/GMLReader.java#L67

and is thus unable to correctly parse a GML geometry from a regular GML document

This also touches upon #144

situx:
Changes a wrong namespace in the GML example as discovered by #330 

Closes #330

### DIFF
--- a/1.1/spec/11-Part-08.adoc
+++ b/1.1/spec/11-Part-08.adoc
@@ -441,7 +441,7 @@ The example <<RDFS Datatype: geo:gmlLiteral, GML Literal>> below encodes a point
 """
 <gml:Point 
         srsName=\"http://www.opengis.net/def/crs/OGC/1.3/CRS84\" 
-        xmlns:gml=\"http://www.opengis.net/ont/gml\">
+        xmlns:gml=\"http://www.opengis.net/gml/3.2\">
     <gml:pos>-83.38 33.95</gml:pos>
 </gml:Point>
 """^^<http://www.opengis.net/ont/geosparql#gmlLiteral>


mperry455 (approved):
Looks good. Thanks Timo.

nicholascar (approved):

PR #344, opened 2022-06-01
Issue 335

### LINKED ISSUE #335: How should Annex B (Functions Summary) be interpreted?
Some things are not made clear yet:

1. What do the 2D and 3D columns mean? Do they mean input or output, or both?
2. How can there be a 3D column when GeoSPARQL does not support 3D geometry?
3. It is not clear where the geometry subtypes are defined. Probably that is SFA-CA. Since the table is normative, does that mean SFA-CA is normative too? Does that mean that all geometry literals should be exclusively modelled on SFA-CA?

### LINKED ISSUE #337: Fix ref [GeoJSON] -> [GEOJSON] in Sec 8.8.3

nicholascar:
See the individual commits to understand how the points raised in #335 are addressed.

As rendered, the table now looks like this:
image

...and explanatory text is added to the Annex' Overview

Closes #335

Closes #337 -  a very small issue


jabhay (approved):
Looks good to me... Some merge conflicts I expected around the headings didn't appear. Just check that these merge well.

PR #350, opened 2022-06-02
Spec corrections and added crossreferences

situx:
Just going through the spec to add some crossreferences and further corrections


jabhay (approved):
Looks good

PR #353, opened 2022-06-15
remove micro credentials; @prefix -> PREFIX

### LINKED ISSUE #352: Remove micro-credentials from ontology
For Issue #22, we decided not to keep attributions at the ontology element level but only at the whole level, so remove any that are like that.

nicholascar:
Closes #352


jabhay (approved):
Looks good to me. Thanks for adding the top level attribution.

Issue #170, opened 2021-06-29
highlight collection membership direction in review notes for implementers


FransKnibbe:
This issue results from [discussion of issue 148](https://github.com/opengeospatial/ogc-geosparql/issues/148#issuecomment-867044138). Membership of SpatialObjectCollections is stated at the collection instance (`rdfs:member`). We should ask implementers if this is OK, or that membership declaration is better done at the member instances (isMemberOf), for example because it allows easier data updates.

dr-shorthair:
If they are not familiar with RDF they will not really understand the question. 
The direction of a triple does not matter. 
The graph is still connected. It can still be queried, updated, sorted either way. 
I really think you are introducing unnecessary complexity here.

FransKnibbe:
Would it be OK if I change it to data store implementers only?

dr-shorthair:
OK

jabhay:
- [ ] @jabhay to determine whether OGC has a preferred or standard way of publishing release/change notes.

jabhay:
From @ogcscotts : there is a template for Release Notes [here](https://github.com/opengeospatial/templates/tree/master/release_notes_template)

jabhay:
- [ ] Create release notes
- [ ] Add question of collection membership direction to release notes for implementers

jabhay:
@jabhay will find an example of one of these for us to use as a guide.

jabhay:
See https://github.com/opengeospatial/ogc-geosparql/pull/359

jabhay:
Still requires this change to be made: 

- [ ] Add question of collection membership direction to release notes for implementers

nicholascar:
Final change - to "Add question of collection membership direction to release notes for implementers" - made in PR #362

PR #368, opened 2022-07-28
Fix Issue 170

nicholascar:
Adds the note requested by @FransKnibbe in Issue #170.

Also makes a number of simple updates to the Release Notes. Itemised in commits.

Closes Issue #170


situx (approved):
All good, but I have two suggestions:
1. Maybe we could add the new class diagram to the release notes
2. The RIF rule script is likely to disappear from GeoSPARQL 1.1, so it should also disappear from the release notes, right?

lvdbrink (approved):

PR #369, opened 2022-07-28
RIF redux

### LINKED ISSUE #361: Conduct external review of RIF files
The RIF files for [1.0](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.0/rules.rifps) & [1.1](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/rules.rifps) were generated a long time ago and probably haven't been reviewed, certainly not by a person external to the SWG who is familiar with RIF. They should be reviewed

nicholascar:
This PR finalises RIF work in GeoSPARQL 1.1 and the result is...

essentially no change to 1.0. This PR removes the script-built RIF-PS files and reverts to keeping a slightly altered version of the RIF template in the Spec.

This is because the script-built RIF file actually hindered understanding of the GeoSPARQL rules, rather than assisting it, due to its length and repetitive nature. Also, the _Presentation Syntax_ used in the file is not specified for rules exchange, only the RIF XML syntax is.

External review of the RIF files indicated that they could be reduced with function encapsulation however complete reduction leads to... just the original template...

A RIF XML expression of GeoSPARQL's rules _could_ be make for GeoSPARQL next version IFF people see merit in their expression in that form.

Thanks @der for your external review!

Closes #361


situx:
I think we should provide such a RIF XML expression. After all, should we not provide a RIF artifact for people to use directly?

nicholascar:
I think I've run out of steam to do a RIF XML form as it requires learning a bit about the RIF formats to expand the template in the Spec into a valid RIF encodings - Presentation Syntax or RDF and for what benefit really? The rules just say make function `x(a, b)` == `a ->x-> b`... Super obvious really...

jabhay (approved):
If I understand this correctly, and through discussion, what we're doing here is rolling back to the original, which makes sense to me.

PR #370, opened 2022-07-28
Editorial citation tweaks

nicholascar:
...and a single missing DGGS IRI added to an example data file.


situx (approved):
All good but I would suggest to change the bibtex key of the citation suggestions, maybe to ogc2012geosparql10 and ogc2022geosparql11

jabhay (approved):
Looks good to me.

PR #371, opened 2022-08-01
replace en dash with hyphen

lvdbrink:
Replaced character U+2013 (en dash) with U+002d (hyphen) to avoid ttl parse problem. A user of I think topquadrant brought this to my attention.


nicholascar (approved):
Merging after one review since it's a single character change!

PR #383, opened 2022-08-25
Fix Non-Top functions listing and SFA comparison

nicholascar:
Ensures we have 

* paired functions with properties
    * e.g. `geof:perimeter` for `geo:hasPerimeter`
* metric/non-metric pairs 
    * e.g. `geof:perimeter` & `geof:metricPerimeter`
* return types matching function
    * e.g. `geof:perimeter`  returns an `rdf:Resource` where as `geof:metricPerimeter` returns an `xsd:double`

Also adds all SFA functions to Table _B.2 GeoSPARQL to SFA Functions Mapping_. We were missing lots.

In your reviews, please indicate what we should do about the un-supported SFA functions.

### DIFF
--- a/1.1/spec/11-Part-08.adoc
+++ b/1.1/spec/11-Part-08.adoc
@@ -869,10 +869,10 @@ This extension mechanism enables GeoSPARQL implementations to simultaneously sup
 
 NOTE: Several non-topological query functions use a unit of measure IRI. See the <<Recommendation for specification of units of measurement,  Recommendation for specification of units of measurement>>. Also, the OGC has recommended units of measure vocabularies for use, see the OGC Definitions Serverfootnote:[https://www.ogc.org/def-server].
 
-==== Function: geof:mericArea
+==== Function: geof:metricArea
 
 ```
-geof:mericArea (geom: ogc:geomLiteral): xsd:double
+geof:metricArea (geom: ogc:geomLiteral): xsd:double
 ```
 
 Returns the area of `geom` in square meters. Must return zero for all geometry types other than Polygon. This function is similar to <<Function: geof:area, `geof:area`>> but does not need a specification of measurement unit.
--- a/1.1/spec/11-Part-08.adoc
+++ b/1.1/spec/11-Part-08.adoc
@@ -880,7 +880,7 @@ Returns the area of `geom` in square meters. Must return zero for all geometry t
 ==== Function: geof:area
 
 ```
-geof:area (geom: ogc:geomLiteral): rdf:Resource
+geof:area (geom: ogc:geomLiteral, units: xsd:anyURI): rdf:Resource
 ```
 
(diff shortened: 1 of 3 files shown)


situx (changes requested):
I have added many corrections, myself. In particular, parameters were missing for geof:area and geof:length and the functions were not included in funcsrules.
However, you changed the function signature of geof:buffer and geof:distance, both GeoSPARQL 1.0 functions and this would not be allowed for backwards compatibility. The rdf:Resource should be reverted to xsd:anyURI. 
As for the Simple Features functions: We have an issue to add them all in at latest GeoSPARQL 1.2 and that is how I would proceed unless we want to add them now.

nicholascar:
> We have an issue to add them all in at latest GeoSPARQL 1.2 and that is how I would proceed unless we want to add them now.

OK, I'll try and wind back as much as I can, but I'll have to preserve a few things that are necessary for correctness.

situx:
I think you do not have to wind back anything. Adding missing functions to the SFA table should be absolutely fine and can be seen by readers as an outlook of what will be added in GeoSPARQL 1.2 onwards. My only concern is that the parameters of functions such as geof:buffer should match GeoSPARQL 1.0 or be more general. However, here you restrict parameters, i.e. xsd:anyURI to rdf:Resource.

1.1/spec/11-Part-08.adoc
```diff
@@ -896,17 +910,17 @@ geof:metricBuffer (geom: ogc:geomLiteral,
...
 
-Returns a geometric object that represents all Points whose distance from `geom` is less than or equal to the `radius` measured in meters. Calculations are in the coordinate reference system of `geom`. This function is similar to <<Function: geof:buffer, `geof:buffer`>>, but does not need a specification of distance unit.
+Returns a geometric object that represents all Points whose distance from `geom` is less than or equal to the `radius` measured in meters. Calculations are in the coordinate reference system of `geom`. This function is similar to <<Function: geof:buffer, `geof:buffer`>>, but does not need a specification of measurement unit.
 
 ==== Function: geof:buffer
 
 ```
 geof:buffer (geom: ogc:geomLiteral, 
-             radius: xsd:double, 
+             radius: rdf:Resource,
```
mperry455, line 919: Why did radius change from xsd:double to rdf:Resource?

nicholascar: I don't know! I think this is a mistake by me. I think I've gone too far typing to implement the pairs of metric/non-metric functions and inadvertently set every input for non-metric functions to objects, not doubles, but this function has a `units` parameter so that should be all that's needed with a double numerical value. I'll change this back.

nicholascar:
I've fixed my mistake for `buffer` and Timo's updates are in place and merge conflicts resolved, so please re-review!

situx (approved):
Looks good to me now and also provides a roadmap of which functions are still left to define until we have included all SFA functions in the next release

PR #389, opened 2022-10-22
Full review of the standard minus annexes

jabhay:
This branch represents a full review of the standard excluding the annexes.

It also incorporates feedback from Karl Reed.


nicholascar (approved):
Good work here Joe!

mperry455 (approved):
Looks good.

PR #405, opened 2023-03-13
Elaborate on length function

### LINKED ISSUE #397: Question about line length
In wake of the upcoming GeoSPARQL 1.1, I want to take the chance to clarify some things (maybe the answer is already clearly in the standard and I just missed it, in which case my apologies)

8.9.25. Function: geof:length (and metricLength)

> The longest length from any one dimension is returned

the wording is a bit unclear to me, does this give me the length of a (Multi)LineString, as in JTS's Length.ofLine, or only the linear distance between its start and end? If the latter, how can I get the length of a line?

situx:
Closes #397 
@SimonBin does the clarification work for you or do you have other suggestions for the wording?


lvdbrink (approved):

SimonBin:
Thank you!

PR #409, opened 2023-03-15
Add perimeter to spec

### LINKED ISSUE #401: perimeter (and metricPerimeter) are not documented
in the upcoming GeoSPARQL 1.1 document as far as I can tell.

they are listed in 8.9 Req 41 and are surely a useful addition

situx:
Closes #401

I also added corrections to the function definitions, so that each function now has a clickable function URI.

One question:
I added the phrase, "If no unit parameter is given, the result is returned in the unit of the spatial reference system"
Should we add that to all other non-metric functions as well, or should I remove it?
We have not talked about optional parameters yet, but it seemed logical to me to add it and comparable implementations like PostGIS also work this way.


lvdbrink (approved):
For me it makes sense to add that comment about the unit in other places where it applies; or perhaps there's a place in the document where it can be added as a general remark for all functions it applies to.

mperry455 (approved):

lvdbrink (approved):

jabhay (approved):
Latest changes cover off on recommendations from reviewers.

Issue #241, opened 2021-12-01
Cater for and define basic Geometry roles


nicholascar:
The role that a Geometry may play with respect to a Feature allows for Features with multiple Geometries that can be known to do different things.

Roles might subsume specialised versions of `hasGeometry`, e.g. `hasCentroid`.

If GeoSPARQL allows roles for Geometries:

* it should supply some (expected) common roles
    * centroid, bounding box etc
* indicate how the role target is defined
    * i.e. a Geometry has a role _with respect to_ what?? A Feature, but what if it's used by multiple Features?
* provide a register of roles for expansion outside the standard development lifecycle

dr-shorthair:
While property-specializations make sense from a semantic viewpoint, they can obfuscate queries. 
So I think we need to be circumspect in coining too many specialized predicates. 
Roles on qualified associations might be a way to go, but that suggests that we would need something like 
```turtle
geo:hasQualifiedGeometry a owl:ObjectProperty ;
    rdfs:range geo:QualifiedGeometry . 

geo:QualifiedGeometry
  rdf:type owl:Class ;
  rdfs:subClassOf [
      rdf:type owl:Restriction ;
      owl:cardinality "1"^^xsd:nonNegativeInteger ;
      owl:onProperty geosparql:hasGeometry ;
    ] ;
  rdfs:subClassOf [
      rdf:type owl:Restriction ;
      owl:cardinality "1"^^xsd:nonNegativeInteger ;
      owl:onProperty geosparql:hasGeometryRole ;
    ] ;
.
```

nicholascar:
> ...suggests that we would need something like...

I'm still concerned about how the Roles are targeted, e.g. for this data:

```
ex:qual-geo-01 
    a geo:QualifiedGeometry ;
    geo:hasGeometry ex:geom-01 ;
    geo:hasGeometryRole ex:role-x ;
.
```

Can this make sense without a Feature? Surely not as the role is relative. So how do we ensure to context of the role? Perhaps we need:

```
ex:geo-01 
    a geo:Geometry ;
    geo:asWKT "..." ;
    geo:hasRelativeRole [
        geo:role ex:centroid ;
        geo:roleFor ex:feature-01 ;
    ] ; 
.
```

So here the role is only defined with respect to `ex:feature-01`. We don't need a qualified Geometry as a normal Geometry can have multiple roles, with respect to multiple Features, defined.

[...]

PR #418, opened 2023-03-29
Draft to include styles for features

### LINKED ISSUE #64: Extending the GeoSPARQL ontology with feature style descriptions
I have developed several tools in the last year which required me to define a feature style for certain geometry representations modelled with GeoSPARQL.
I ended up defining:

- A style property which links a geo:Geometry to a style description instance with
  - point/lineString/Polygon style properties with CSS literals
  - pointImage/lineStringImage/polygonImage properties with URIs pointing to Icons or SVG definitions (SVG literals)
  - hatch: Whether to draw the geometry in a hatched style
  - lineStringImageStyle: How the image behaves in the lineString (how often it is repeated etc.)

There may be several style representations per geometry and there could be constraints attached to each of these styles which link to certain semantic classes (for now modelled as text literals by me).
So there could be the possibility for reasoning eligible styles for a Geometry representation according to the data given.

What I currently do with these style representations is to downlift them to GeoJSONCSS (https://github.com/albburtsev/Leaflet.geojsonCSS) and then add them to a Leaflet view.

My questions are therefore:

- Is the GeoSPARQL ontology the right place to standardize feature style descriptions?
- Is there any other related work known to members of the group or best practices I have overlooked?

situx:
Closes #64 

This draft is loosely based on the GeoJSONCSS approach:
https://wiki.openstreetmap.org/wiki/Geojson_CSS

It introduces a class for Styles, and properties to connect these to SpatialObjects and literals to store style information.

Ideas:
* Conditions for styles could be enforced with SHACL rules
* Linking geo styles to geometry roles that @nicholascar is proposing in #241


situx:
This draft is mainly complete, but an open question for me is still the definition and positioning of text.
Should it be a label from the feature?
How to signal that a label should be there and where it should be?

nicholascar:
I like the idea of a GeoSPARQL-aligned Semantic Web handling of styles for features but I don't think it's core to GeoSPARQL so shouldn't be included in GeoSPARQL itself but rather an extension ontology (that could be published by OGC). This would be similar to the Fuzzy Geometries model: interesting but not core.

PR #421, opened 2023-04-05
Incorporating changes to ConcaveHull description in specification

### LINKED ISSUE #270: Clarification needed for SFA Geometry class
While translating the English texts in the [Simple Features Vocabulary](https://github.com/opengeospatial/ogc-geosparql/blob/master/1.1/sf_geometries.ttl) I came across a bit of text that I do not understand. This is part of the explanation of the `Geometry` class (in both the SFA document and the vocabulary):

> All Geometry classes described in this standard are defined so that instances of Geometry are topologically
> closed, i.e. all represented geometries include their boundary as point sets. This does not affect their
> representation, and open version of the same classes may be used in other circumstances, such as topological
> representations.

It is the second sentence that I do not understand. The part "and open version of the same classes" looks grammatically incorrect to me. Probably it should read "open versions". 

I understand representations to mean encodings like WKT and WKB. So how does Geometry instances being topologically closed not effect their representations? 

And how could an open version of a geometry class be used in topological representations?

jabhay:
We're only handling the case of a ConcaveHull without any parameters because various implementers use different parameters and trying to handle all permutations of these is outside of the scope of the standard. As such, implementers will elect to set default values, and we wish to ensure that these are documented by them for their users.

Fixes #270

### DIFF
--- a/1.1/spec/11-Part-08.adoc
+++ b/1.1/spec/11-Part-08.adoc
@@ -947,7 +947,7 @@ The function http://www.opengis.net/def/function/geosparql/convexHull[`geof:conv
 geof:concaveHull (geom: ogc:geomLiteral): ogc:geomLiteral
 ```
 
-The function http://www.opengis.net/def/function/geosparql/concaveHull[`geof:concaveHull`] returns a geometric object that represents all Points in the concave hull of `geom`. Calculations are in the spatial reference system of `geom`.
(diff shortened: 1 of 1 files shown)


mperry455 (approved):

situx (approved):
Seems good to me. At least it reflects what we discussed in previous meetings

lvdbrink (approved):

PR #422, opened 2023-04-05
Fixing typo in Simple Features spec in our corresponding vocab

jabhay:
Responds to #270 but doesn't fix the underlying issue.


situx (approved):
Correction of the typo seems good to me

mperry455 (approved):

jabhay:
Bringing this one through as the solution was recommended by Frans already - hence I feel it has the requisite three agreements - implementor and two others.

PR #443, opened 2023-10-06
Draft for inclusion of surface textures and materials

### LINKED ISSUE #438: Standardization of textures for geometries
### Observation

An increasing number of linked data users are building 3D Digital Twins with GeoSPARQL. In a 3D Digital Twin, the textures that are displayed on top of the geometries are often important. There are linked data standards about storing images and there is GeoSPARQL for storing (3D) geometries. But there are enough challenges in combining the two that require further standardization.

### Expected

Standardization of the way in which textures and geometries should be optimally combined. This may include how the texture is positioned on top of the geometry, and how the texture and geometry are related (e.g. `geo:hasTexture`).

### Additional remarks

I am supplier of the TriplyDB linked data product. As a supplier, I receive the question above increasingly often, so I can detect the demand for standardization in this area. I am not a geo/3D expert, so I do not know how standardization of this issue should look like. (But if this would be standardized, then I would be able to act as one of the first implementors.)

situx:
This pull request proposes to include textures and materials by examining related work such as CityGML/CityJSON, among others.
Finishing this pull request will close #438

The following contents are defined:

* Classes and properties for Materials and Textures 
* Connection of surfaces to textures and materials (This part depends on the definition of 3D geometries in GeoSPARQL, which will be a separate pull request)

TODO:
* Evaluating whether literals need to be defined to store textures in the knowledge graph
* Cross-checking with the pull request for geometry styles in 2D #418


nicholascar:
I like the idea of a GeoSPARQL-aligned Semantic Web handling textures and materials but I don't think it's core to GeoSPARQL so shouldn't be included in GeoSPARQL itself but rather an extension ontology (that could be published by OGC). This would be similar to the Fuzzy Geometries model: interesting but not core.

situx:
@nicholascar agreed, it should not be core.
I think either, we might used the appearance module of CityRDF or we define an appearance vocabulary ourselves.
Given that we also need a vocabulary for 3D primitives, I think the fate of this pull request might depend on our discussion of #611

PR #469, opened 2023-12-22
Function parameter registry work

situx:
Adaptations to functions.ttl to create a function parameter registry:

- [x] Proposal of a new namespace for function parameters
- [ ] Question: Do we need to replace blank nodes in parameter list definitions with actual URIs?
- [ ] Define the concept scheme for function parameters


nicholascar:
This PR is confusing since so many files are changes, I assume for repo management reasons and not due to the core alterations of the PR.

Can this please be re-issued against the latest branches so there are few changes?

situx:
I believe we adopted the new namespace for function parameters and output elsewhere already. I believe this pull request is superseded and can be closed without merging

nicholascar:
I think this PR should be abandoned:

* most of the substantive changes have been made elsewhere
* there are too many conflicts with the very different repo structure, since 2023, to realistically merge

We just need to check if there are any important changes not applied elsewhere.

PR #502, opened 2024-02-10
VANN Namespace prefixes for GeoSPARQL

situx:
I added Namespace prefix statements from the vann vocabulary to the GeoSPARQL vocabulary and simple features vocabulary.

- [x] Added statements to geo and SF vocabulary
- [x] Should we do that for the rules and functions namespace as well and do we need to declare them a voaf:Vocabulary for that?
- [x] The namespace gsp is registered at [Linked Open Vocabularies](https://lov.linkeddata.es/dataset/lov/vocabs/gsp) for the GeoSPARQL vocabulary. Is this the prefix we would like to recommend or would we rather recommend geo, which is currently used to described the [W3C Geo Vocabulary](https://lov.linkeddata.es/dataset/lov/vocabs/geo) ?


mperry455:
Based on discussion in 02/21/24 meeting, the group would prefer to use geo as the recommended prefix.

vocabularies/functions.ttl
```diff
@@ -17,6 +17,8 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 cs:
     a skos:ConceptScheme ;
     dcterms:created "2021-06-16"^^xsd:date ;
+    vann:preferredNamespacePrefix "geof"^^xsd:string ;
```
mperry455, line 20: Don't we need a new PREFIX entry for vann

situx: Correct. I forgot it in this particular file. I added it now

vocabularies/sf_geometries.ttl
```diff
@@ -18,6 +19,8 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
         ] ;
     dcterms:license "https://www.ogc.org/license"^^xsd:anyURI ;
     dcterms:modified "2023-11-30"^^xsd:date ;
+    vann:preferredNamespacePrefix "sf"^^xsd:string ;
+    vann:preferredNamespaceUri "http://www.opengis.net/ont/sf"^^xsd:anyURI ;
```
SimonBin, line 23: missing # ?

situx:
We agreed to adopt the VANN namespaces discussed in this pull request and I added them to every branch in every geo.ttl and sf_geometries.ttl file. The pull request therefore does not have to be merged.

Issue #160, opened 2021-06-15
Function to retrieve lat/lon


situx:
I just stumbled upon this StackOverflow post from many years ago asking how to retrieve lat/long from a GeoSPARQL-encoded geometry.
https://stackoverflow.com/questions/22536467/how-to-retrieve-latitude-and-longitude-from-geosparqls-wktliteral
The solution seems rather cumbersome.
Now with GeoSPARQL 1.1 we could (mis-)use e.g. the maxX() and maxY() function to retrieve lat/long from a single point, but it really seems like a fundamental thing to do that is currently not well supported in GeoSPARQL.
Any opinions about this?

FransKnibbe:
I think minX() and maxY() should be sufficient if there is a clear and easy way to find out what X and Y mean. So this again shows the need for a CRS ontology and CRS registry based on that CRS ontology.

FransKnibbe:
Also, I think it make sense to allow defining geometries directly in RDF. At least for point locations it would make sense. See also [issue 42](https://github.com/opengeospatial/ogc-geosparql/issues/42).

jabhay:
A "solution" exists in 1.1. Shifting this to proposed scope for 1.2.

SimonBin:
> I think maxX() and maxY() should be sufficient

I disagree because of the rest of your sentence, it would be very convenient to have lat and lon accessors. This is not a "solution"

We have implemented this as an extension to Apache Jena as `geof:lat` and `lon` and can recommend such functions to be added: https://github.com/Scaseco/jenax/blob/jena-4.8.0/jenax-arq-parent/jenax-arq-plugins-parent/jenax-arq-plugins-bundle/src/main/java/org/aksw/jena_sparql_api/sparql/ext/geosparql/GeoSparqlExFunctions.java#L133

situx:
We merged #526 recently which propose a solution for this issue by defining the geof:easting and geof:northing functions. So this can be closed

Issue #398, opened 2023-02-13
Question about geometry function calculations and SRS


SimonBin:
In wake of the upcoming GeoSPARQL 1.1, I want to take the chance to clarify some things (maybe the answer is already clearly in the standard and I just missed it, in which case my apologies)

8.9.1

> An invocation of any of the following functions with invalid arguments produces an error. An invalid argument includes any of the following:
> 
>     A geometry literal from a spatial reference system that is incompatible with the spatial reference system used for calculations

can the `geof:buffer` function add a metric buffer around a line in CRS84?

(some background: in Apache Jena, this calculation currently fails with an error, because CRS84 is not a metric SRS)

I see there is a new `geof:metricBuffer`, presumably this function would be able to do so in any case?

situx:
The only difference between the geof:metricBuffer and the geof:buffer function is that the unit is that the distance in geof:metricBuffer is fixed to meters. (this is thee same for all "metric" functions in GeoSPARQL 1.1)
However, you may as well use the function geof:buffer for a metric buffer, as long as you give it a metric unit as a parameter.

Both functions return a Geometry representation e.g., a WKT Polygon, which represents the buffer area in the CRS of the input geometry literal.
Both functions are also not restricted by Geometry type. So defining a buffer around a LineString is a valid operation.

The output of the function geof:buffer is intentionally not more precisely defined in GeoSPARQL 1.0 and 1.1. So you could return any literal type. In fact, triple stores returned very different things for buffer, as we found out in the GeoSPARQL 1.0 compliance benchmark.

SimonBin:
So according to your opinion, a metric unit is _not_ an invalid argument and the triple store is required(?) or suggested(?) to do the conversion->metric->calculate->spheric, even though obviously it's not possible to do a metric calculation on CRS84 directly?

situx:
Ultimately that is a task up to the implementers to decide. It is also valid to say that a conversion is impossible, as the unit does not correspond to a unit defined in the used CRS, and to throw an exception. 
You might say that this is underspecified, and you might be right. 

However, this is likely nothing we can fix in GeoSPARQL 1.1, as we need to be backward compatible with GeoSPARQL 1.0 where the geof:buffer function was initially defined. It might be in scope for a GeoSPARQL 1.2 iteration where we make more substantial changes to the standard.

VladimirAlexiev:
[...]

PR #526, opened 2024-06-07
Latitude and Longitude Functions

situx:
As suggested in #160 I am proposing the inclusion of a latitude and a longitude function for GeoSPARQL.

We may have two possibilities for definition here:
1. Define the function for points only (which was also implemented in the Apache Jena extension)
2. Define the function for all geometries. For geometries other than points find a solution as to which latitude/longitude to return. My current proposal says "return the latitude/longitude of the centroid of the given geometry" but that is just one possible proposal.

@nicholascar and @SimonBin any opinions?


SimonBin:
I think doing 1. is fine, as for prior art we have e.g. MySQL: https://dev.mysql.com/doc/refman/8.4/en/gis-point-property-functions.html#function_st-latitude 

if the user desires the outcome of 2., they can call geof:centroid manually

spec/sections/11-geometry_extension.adoc
```diff
@@ -1632,6 +1634,23 @@ The function http://www.opengis.net/def/function/geosparql/length[`geof:length`]
 See the <<Recommendation for units of measure, Recommendation for units of measure>>.
 ====
 
+[%unnumbered]
+[source,turtle]
+----
+geof:latitude (geom: ogc:geomLiteral): xsd:double
+----
+
+The function http://www.opengis.net/def/function/geosparql/latitude[`geof:longitude`] returns the latitude of `geom` in meters. If the geometry is a point it returns the point's latitude. If the geometry is a more complex geometry it returns the latitude of its centroid as calculated by geof:centroid.
```
SimonBin, line 1643: in metrers -> in degree?

situx: How about "returns the latitude of `geom` in the unit that is proposed by the coordinate reference system associated with the geometry"?

SimonBin: from my feeble understanding, lat and lon would only ever be defined as an angle (normal to the surface and plane of the celestial body resp. rotational angle between the zero meridian), and thus only degree (or radian??) would be suitable. The other question is related to #398 i.e. should a point defined on a metric (projected) CRS be eligible for automatic transformation into CRS:84 in order to determine its lat/lon on the celestial body or not

jabhay:
Discussion in the group:

Should this be for points only? Options presented were:

1. Define the function for points only (which was also implemented in the Apache Jena extension)
2. Define the function for all geometries. For geometries other than points find a solution as to which latitude/longitude to return. My current proposal says "return the latitude/longitude of the centroid of the given geometry" but that is just one possible proposal.

Option 1 was selected as there is already a function provided to get the centroid of a shape other than a point. We don't want to make assumptions over how to chain this together to produce a result for non-Points given the functions required for the user to decide that are available.

How should the centroid be returned? Options presented:

1. Always a particular unit.
2. In the unit of the CRS.

Option 2 was selected as it is aligned with how Simple Features functions work.

nicholascar:
1. Please add the function definition to the [B.1.  Functions Summary Table](https://docs.ogc.org/is/22-047r1/22-047r1.html#_8646c885-ac65-42bf-bfa9-8d170fa7b4e9)
2. Please provide an example in [Annex C - GeoSPARQL Examples](https://docs.ogc.org/is/22-047r1/22-047r1.html#_3e570415-5147-4114-a8f9-6dcda569c0fd)

situx:
Meeting 26th June: We decided to rename the functions to geof:easting and geof:northing so that values in meter may be returned

situx:
- Updated this pull request to use geof:easting and geof:northing
- Added an example in Annex C
- Added the functions to the Functions table
- Added a file that reprents the RDF used in Annex C to the examples folder

nicholascar (approved):
Just curious why the example data is in RDF/XML as opposed to Turtle??

situx:
I agree, we could add a TTL version. It is RDF/XML because in the GeoSPARQL 1.0 spec in the Annex the example is also RDF/XML. In the GeoSPARQL Compliance Benchmark we also chose to stick to RDF/XML for consistency

jabhay (approved):
Looks like what we discussed has been implemented.

situx:
I added the TTL version of the example from Annex C. With that I will now merge this pull request.

PR #533, opened 2024-07-11
Change empty namespace in geo.ttl to geo:

### LINKED ISSUE #474: don't use BASE and empty prefix; precise ontology URL; use turtle `@prefix` syntax
The ontology https://opengeospatial.github.io/ogc-geosparql/geosparql11/geo.ttl has this preamble:
```
BASE <http://www.opengis.net/ont/geosparql>

PREFIX : <http://www.opengis.net/ont/geosparql#>
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX sdo: <https://schema.org/>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX spec11: <http://www.opengis.net/spec/geosparql/1.1/specification.html#> # TODO: use PID IRI
```
The ontology URL is
```ttl
<http://www.opengis.net/ont/geosparql> 
	a owl:Ontology ;
	dcterms:title "GeoSPARQL Ontology" ;
```
And then each term has `rdfs:isDefinedBy :` .

Suggestions:
- there are no relative URLs in the file, so don't use `BASE`
- don't use an empty prefix. Many ontologies do that (including SKOS) but I think it's a bad practice. Many repos (including GraphDB) record incoming Turtle prefixes in their namespace registries, but there can only be one empty prefix, so it will be appropriated by the first ontology to be loaded. 
- do declare the preferred prefix and namespace using VANN. According to prefix.cc (https://prefix.cc/geo), the most popular prefix for GeoSPARQL is `geo:`
- use the same URL in the ontology declaration and in `rdfs:isDefinedBy`. Currently the two differ since the latter has an extra `#` appended. This means that when a web crawler stumbles upon a GeoSPARQL term, it cannot discover the whole GeoSPARQL ontology.
- use the usual Turtle syntax `@prefix`. It is true that Turtle now also allows the SPARQL syntax `PREFIX` but IMHO using the usual syntax makes the ontology slightly more interoperable.

In summary, I propose to change it as follows:

```ttl
@prefix geo: <http://www.opengis.net/ont/geosparql#>.
@prefix vann: <http://purl.org/vocab/vann/>.
@prefix dcterms: <http://purl.org/dc/terms/>.
...

geo:
	a owl:Ontology ;
	dcterms:title "GeoSPARQL Ontology" ;
	vann: preferredNamespacePrefix "geo";
	vann:preferredNamespaceUri "http://www.opengis.net/ont/geosparql#";
...
<term> rdfs:isDefinedBy geo: ;
```

### LINKED ISSUE #509: `geo.ttl` should not use the empty prefix
`geo.ttl`:
```
PREFIX : <http://www.opengis.net/ont/geosparql#>
```
However, the spec and other turtle files use the prefix `geo:` for this.

Using an empty prefix is IMHO a bad practice (started as early as SKOS) since:
- It sets a bad example: there's only one empty prefix, so clients cannot use empty prefix in their own data
- Semantic repositories often use prefixes defined in ingested files as "namespaces" to be used in prefix autocompletion in SPARQL. But you would not want GeoSPARQL to be associated with `:`
- It makes search in the ontology file harder. In fact this issue was mis-described as "geo:SpatialObject is not defined in `geo.ttl` or any of the other ontologies" because of this problem.

In addition, GeoSPARQL should declare its namespace using VANN:
```ttl
<http://www.opengis.net/ont/geosparql> 
  a owl:Ontology ;
  vann:preferredNamespacePrefix  "geo" ;
  vann:preferredNamespaceUri     "http://www.opengis.net/ont/geosparql#" ;
```

situx:
Closes #509 
Closes #474

Issues to clarify:

- [ ] Change ontology URI to include an extra # ?
- [x] Choose TTL syntax or SPARQL syntax in geo.ttl file: It would seem we want to stay with the SPARQL syntax here

### DIFF
--- a/vocabularies/geo.ttl
+++ b/vocabularies/geo.ttl
@@ -2,7 +2,7 @@
 
 BASE <http://www.opengis.net/ont/geosparql>
 
-PREFIX : <http://www.opengis.net/ont/geosparql#>
+PREFIX geo: <http://www.opengis.net/ont/geosparql#>
 PREFIX dcterms: <http://purl.org/dc/terms/>
 PREFIX owl: <http://www.w3.org/2002/07/owl#>
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
@@ -18,7 +18,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 	a owl:Ontology ;
 	dcterms:title "GeoSPARQL Ontology" ;
  	vann:preferredNamespacePrefix "geo" ;
-    vann:preferredNamespaceUri "http://www.opengis.net/ont/geosparql#" ;
+        vann:preferredNamespaceUri "http://www.opengis.net/ont/geosparql#" ;
 	dcterms:publisher [
 		a sdo:Organization ;
 		sdo:name "Open Geospatial Consortium" ;
@@ -99,7 +99,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 # #
 # #################################################################
 
-:gmlLiteral
+geo:gmlLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy
 		: , 
@@ -112,7 +112,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 		spec11:C.1.2.4 ;
 .
 
-:wktLiteral
+geo:wktLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy 
 		: ,
@@ -125,7 +125,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 		spec11:C.2.2.2 ;
 .
 
-:ewkbLiteral
+geo:ewkbLiteral
 	a rdfs:Datatype ;
 	rdfs:isDefinedBy 
 		: ,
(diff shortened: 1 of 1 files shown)


nicholascar (approved):

nicholascar:
> Choose TTL syntax or SPARQL syntax in geo.ttl file

I much, much _much_ prefer `REFIX` over `@prefix` as then the Turtle corresponds better with SPARQL and prefix statements can be copied between SPARQL and Turtle.

The [Turtle 1.1 note on this](https://www.w3.org/TR/turtle/#h_note_1) says:

"It is advisable to serialize RDF using the '@prefix' and '@base' forms until RDF 1.1 Turtle parsers are widely deployed."

Well, RDF 1.1 is widely deployed now and we are almost at RDF 1.2, so we should be using `PREFIX`.

I'm happy to drop the `BASE` directive from the ontology, once this PR is merged as it will no longer be in use.

situx:
Good. I removed the BASE directive. Now the only unclear thing is whether the base URL should include a # or not

nicholascar (approved):

mperry455 (approved):
Looks good.

Issue #110, opened 2021-03-12
Should transformations between spatial reference systems be supported?


FransKnibbe:
A question that arose at the 2021-03-10 teleconference: Should GeoSPARQL have a function to allow transformation of geometries from one spatial reference system (SRS) to another? Such a function is available in many GIS software ([PostGIS for example](https://postgis.net/docs/ST_Transform.html)).

FransKnibbe:
My thoughts: SRS transformation is a very useful function for many applications. But I think it does need a framework for defining spatial reference systems. A coordinate transformation function needs to be able to access the required SRS parameters. Assuming we prefer SRSs to be identified and accessed by IRI, an SRS transformation function would require an SRS ontology (see [issue 12](12)).

situx:
I have created a draft pull request with some suggestions for this #108 and a proposal on whether we should also support transformations between literal formats. I looked at POSTGIS as an example.
Two common function signatures exist in POSTGIS from which we could infer:
geof:transform(geom: ogc:geomLiteral, from: xsd:anyURI, to: xsd:anyURI) ogc:geomLiteral
geof:transform(geom: ogc:geomLiteral, to: xsd:anyURI) ogc:geomLiteral
The first requires the source SRS to be stated, the second retrieves it from the literal (or possibly) another property in the background for GeoSPARQL 2.0.
In my opinion the second signature might be sufficient, the first signature might be interesting if we were to define geometry literals without a SRS attached.

I second the need for a SRS ontology, but for GeoSPARQL 1.1 we could stick to the definition of a SRS as an IRI as it is currently done in WKT and GML literals and rely on the triplestore implementation to provide their own database to resolve the SRS parameters.

mathib:
@situx the second function is interesting and useful indeed for geometries with an unknown SRS!

situx:
> @situx the second function is interesting and useful indeed for geometries with an unknown SRS!

You are right but I think using GeoSPARQL 1.0 the CRS cannot be undefined or unknown in all literal types as we established before.

So maybe one could add the second signature when it is possible to define those in the standard

jabhay:
Adding a second signature to the standard sounds like something we'd like to do, but probably this function is not alone. A good candidate for version 1.2. I'll add it to the proposed features for that version.

PR #534, opened 2024-07-12
Default parameter for transform function

### LINKED ISSUE #503: Default value for second argument in geof:transform()
Hi GeoSPARQL friends!,

It is great that GeoSPARQL 1.1 will include function [1].

```
[1] geof:transform (geom: ogc:geomLiteral, srsIRI: xsd:anyURI): ogc:geomLiteral
```

Since transformations to CRS [2] are particularly common, would it make sense to make the second argument to function [1] optional? If the CRS is absent, CRS [2] is used by default.

```
[2] http://www.opengis.net/def/crs/OGC/1.3/CRS84
```

This is consistent with [2] already being the default CRS in WKT literals.

situx:
Closes #503

This pull request attempts to define a default parameter in the transform function.

My main question here is:

Is that the way you define a default parameter in the spec?
Or is the correct way to do that to define two function definitions?

If we define default parameters like this in the future, is that understandable enough for implementers?

Also; A good idea to revisit #110 in this discussion

### DIFF
--- a/vocabularies/functions.ttl
+++ b/vocabularies/functions.ttl
@@ -420,7 +420,8 @@ geofp:transform_param1 a fno:Parameter ;
 
 geofp:transform_param2 a fno:Parameter ;
     fno:type xsd:anyURI ;
-    fno:required "true"^^xsd:boolean ;
+    fno:required "false"^^xsd:boolean ;
+    rdf:value "http://www.opengis.net/def/crs/OGC/1.3/CRS84"^^xsd:anyURI ;
     skos:prefLabel "srsIRI"@en .
 
 geofo:transform_output a fno:Output ;
--- a/vocabularies/functions.ttl
+++ b/vocabularies/functions.ttl
@@ -420,7 +420,7 @@ geofp:transform_param1 a fno:Parameter ;
 
 geofp:transform_param2 a fno:Parameter ;
     fno:type xsd:anyURI ;
-    fno:required "false"^^xsd:boolean ;
+    fno:required "true"^^xsd:boolean ;
     skos:prefLabel "srsIRI"@en .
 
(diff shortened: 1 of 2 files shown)


mperry455:
Consensus from 7/24 SWG meeting is to revisit set of functions and see where we could apply default parameters.

situx:
@nicholascar 
One remark on default parameters for the non-metric functions:
If we define default parameters there we would be recommending a particular set of URIs to be used to represent e.g. metre.
We avoided such a recommendation in the spec so far.
Maybe we should keep these as they are?

jabhay:
The search to identify functions that meet the criteria above yielded only the Transform function. The remaining question is whether it's worth introducing this feature for a single function? Consensus in the group is to handle this through a function override with one less parameter and a name that indicates the default.

situx:
I added a function definition geof:transformCRS84 which acts as the transform functions with the second parameter equal to CRS84.
The pull request can be reviewed again.

vocabularies/functions.ttl
```diff
@@ -420,13 +420,37 @@ geofp:transform_param1 a fno:Parameter ;
 
 geofp:transform_param2 a fno:Parameter ;
     fno:type xsd:anyURI ;
-    fno:required "true"^^xsd:boolean ;
+    fno:required "false"^^xsd:boolean ;
```
mperry455, line 423: This should be true, correct? Since we decided to use a named function instead of a default parameter value.

situx: Thx. It is now corrected.

mperry455 (approved):

jabhay (approved):
Looks good to me, and lines up with our discussion.

PR #535, opened 2024-07-12
Intersection Geometry Vocabulary

### LINKED ISSUE #239: Provide for topology relations between features with explicit geometries at intersections.
Features that intersect with geo:touches (or other topological relationships) will have an intersection geometry that may or may not be part of the composite geometry of the features - eg. two polygons sharings a LineSegment, two solids sharing a face denoted by a linear ring.

the requirements are:
1) topological statements can be made without defining geometries (supported currently)
2) intersections are features that may have calculated geometries and metadata about how that was done,
3) intersection geometries may reference internal components (such as points) of the feature geometries by id.
4) intersection features will have constraints about the allowable geometries - e.g. a geo:Face would require a gml:LinearRing (?) 
5) intersection features should be consistent with ISO19017 topology complexes, but may be a subset required for "simple features" including its  existing "solid geometry" capabilities.

It is expected to be out of scope for GeoSPARQL to support calculation of these intersections or validation that internal ids are self-consistent, but if the information is expressed an implementation can perform additional integrity checking as desired.

situx:
### This PR has been superseded by #702.

Closes #239

This pull request wants to collect the necessary changes for describing intersection geometries and their properties which were requested in #239

Idea Sketch:

- New SubClass of Geometry: CompoundGeometry
- New SubClass of CompoundGeometry: IntersectionGeometry
- CompoundGeometry geo;hasPart Geometry to describe parts of Compound Geometries

- [ ] How to describe constraints of Intersection Features/Geometries using SHACL?
- [ ] To Discuss: intersection features should be consistent with ISO19017 topology complexes

### DIFF
--- a/vocabularies/geo.ttl
+++ b/vocabularies/geo.ttl
@@ -1050,6 +1050,20 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 		spec11:C.1.1.6 ;
 .
 
+:Intersection
+	a rdfs:Class, owl:Class ;
+	rdfs:subClassOf :Feature ;
+	skos:definition """A feature which is comprised of the fractions of at least two other features"""@en ;
+	skos:prefLabel "Intersection Feature"@en ;
+.
+
+:IntersectionGeometry
+	a rdfs:Class, owl:Class ;
+	rdfs:subClassOf :Geometry ;
+	skos:definition """A geometry which comprises the intersection of at least two other different geometries"""@en ;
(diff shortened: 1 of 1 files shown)


mperry455:
Consensus in 7/24 SWG Meeting is to look at a convention for using provenance to describe that a geometry was derived from an intersection rather than defining new subclasses of geometry.

nicholascar:
Nick to add in a proposal for the provenance

PR #537, opened 2024-07-16
Geocoding Vocabulary

### LINKED ISSUE #444: Geocoding Support in GeoSPARQL
Our group has already received a request from @paulc-dstl to add GeoHash literals to GeoSPARQL.

However, there are more geocoding systems which could be taken into account here.

In fact, a lot of geocoding systems can be resolved to coordinates in either UTM or WGS84-based coordinate reference systems.

My question here:

Would it make sense to maybe define two literal types: "wgs84GeocodingLiteral" and "utmGeoCodingLiteral" and indicate the Geocoding system with a URI definition, similar to how we do it in WKT?
We could curate a list of URIs for well-known geocoding systems which implementers could use.

Or should we proceed with defining GeoHash, and people should create their own literal types outside the standard?

situx:
Closes #444 

This pull request collects URIs and classes for Geocoding Systems which should be registered at the OGC Naming Authority.

The URIs defined in this vocabulary should eventually after review by OGC be used in GeoCode Literals.

### DIFF
--- a/vocabularies/geocoding.ttl
+++ b/vocabularies/geocoding.ttl
@@ -23,6 +23,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
     rdfs:label "Open Location Code"@en ;
     schema1:citation "https://maps.google.com/pluscodes/"^^xsd:anyURI ;
     skos:altLabel "Plus Code"@en ;
+    skos:definition "The Open Location Code is a geocode based on a system of regular grids. It may be used to identify any location on Earth." ;
 .
 
 <http://opengis.net/ont/geocode/What3Words>
--- a/vocabularies/geocoding.ttl
+++ b/vocabularies/geocoding.ttl
@@ -151,6 +151,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
         owl:NamedIndividual ;
     rdfs:label "MGRS"@en ;
     skos:altLabel "Military Grid Reference System"@en ;
+    skos:definition "The geocoordinate standard derived from the UTM grid used by NATO militaries to located points on Earth." ;
 .
 
 :MaidenheadLocatorSystem
--- a/vocabularies/geocoding.ttl
+++ b/vocabularies/geocoding.ttl
@@ -140,6 +140,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
         :TelephoneCode ,
         owl:NamedIndividual ;
     rdfs:label "ITU-T Country Calling Code"@en ;
+    skos:definition "Country calling codes are telephone number prefixes for reaching telephone subscribers in foreign countries or territories" ;
 .
 
 :ITU-TMobileCallingCode
--- a/vocabularies/geocoding.ttl
+++ b/vocabularies/geocoding.ttl
@@ -47,7 +47,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 <http://opengis.net/ont/geocode/OpenLocationCode>
     a
         <http://opengis.net/ont/geocode/GeoCode> ,
-        skos:Concept ;
+        skos:Concept, rdfs:Datatype ;
     skos:prefLabel "Open Location Code"@en ;
     schema1:citation "https://maps.google.com/pluscodes/"^^xsd:anyURI ;
     skos:altLabel "Plus Code"@en ;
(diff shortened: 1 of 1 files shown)


nicholascar:
Can this vocabulary be:

- [ ] proposed not as a class hierarchy but a SKOS ConceptScheme 
    - [ ] like other GeoSPARQL support vocabs with appropriate metadata
    - [ ] with Concepts dual classed as `rdfs:Datatype` objects also
    - [ ] using GeoCode as the parent with others all `skos:narrower`
- [ ] in longturtle RDF format
    * for better Git tracking
    * to remove Protege cruft
    * use https://tools.dev.kurrawong.ai/convert
- [ ] use `schema:citation` instead of `rdfs:isDefinedBy` as the targets are non-RDF nodes
- [ ] provide at least a single-line description for each one
- [ ] using Register Ontology Status Codes for the ConceptScheme and each Concept
    - so they are ready for management of additions, deprecations etc.

I'm listing all these points to see if you all agree to them but I'm not suggesting you (@situx) have to do them all yourself! You've already done the main work here, so I'm happy to do many of these, if you agree to them

situx:
Hi Nick,

I agree with all of your points. I only have a question concerning the SKOS decision:
- What would be the benefit in defining it as as SKOS concept hierarchy? The ability to define it also as rdf:Datatype? Or does it integrate better with OGC Rainbow? I am asking since we also did not define the Simple Features vocab as SKOS and we would also not define the CRS ontology as such.

vocabularies/geocoding.ttl
```diff
@@ -6,8 +6,9 @@ PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
 PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 
 <http://opengis.net/ont/geocode/>
-    a owl:Ontology ;
+    a owl:Ontology, skos:ConceptScheme ;
```
VladimirAlexiev, line 9: if it has no members that are skos:Concept and point to it using skos:inScheme, 
then it's not a ConceptScheme

VladimirAlexiev (changes requested):

vocabularies/geocoding.ttl
```diff
@@ -14,6 +14,7 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
 <http://opengis.net/ont/geocode/GeoHash>
     a
         <http://opengis.net/ont/geocode/GeoHashCode> ,
+        skos:inScheme <http://opengis.net/ont/geocode/GeoCodes> ;
```
VladimirAlexiev, line 17: @situx you inserted "inScheme" between the two RDF types, so now the syntax is invalid.

VladimirAlexiev (changes requested):

situx:
@nicholascar I think I have done most of the changes you requested here.
Could you check and correct what you think is missing?
Then we could get this pull request merged next time.

vocabularies/geocoding.ttl
```diff
@@ -0,0 +1,400 @@
...
+.
+
+:CLLICode
+    a
+        <http://opengis.net/ont/geocode/GeoCode> ,
+        skos:Concept, rdfs:Datatype ;
+    skos:prefLabel "CLLI Code"@en ;
+    skos:inScheme <http://opengis.net/ont/geocode/GeoCodes> ;
+    skos:altLabel "Common Language Location Identifier Code"@en ;
+    skos:definition "A location identifier code used within the North Amerian telecommunications industry to specify the location and function to telecommunications equipment" ;
```
mperry455, line 87: Type "Amerian" -> "American"

vocabularies/geocoding.ttl
```diff
@@ -0,0 +1,400 @@
...
+    skos:example "E09000011 for Royal Borough for Greenwich" ;
+.
+
+:GeoHash-36
+    a
+        <http://opengis.net/ont/geocode/GeoHashCode> ,
+        skos:Concept, rdfs:Datatype ;
+    skos:prefLabel "GeoHash-36"@en ;
+    skos:inScheme <http://opengis.net/ont/geocode/GeoCodes> ;
+    schema1:citation "https://web.archive.org/web/20121227195936/http://www.openpostcode.org/other-codes/geohash-36"^^xsd:anyURI ;
```
mperry455, line 127: Strange looking URI here. Is this correct?

mperry455 (approved):
Looks good. Just marked a couple of typos.

jabhay:
@VladimirAlexiev , could you pls take a look at the final product and approve or comment.

nicholascar (approved):

Issue #590, opened 2024-11-05
Triply Use Case 8: GeoSPARQL profile for IFC


wouterbeek:
# Triply Use Case 8: GeoSPARQL profile for IFC

## Description

A lot of 3D data is stored in IFC. In the future, GeoSPARQL will have the ability to detect or configure the following aspects of 3D BIM data:

- The position, see #586 .
- The height, see #587 .
- The angle, see #588 .
- The size, see #589 .
- The texture, see #584 .

IFCs are able to encode this data, but many IFC files do not include all this information. GeoSPARQL should include a profile for IFC, that allows users to quickly check whether an IFC is eligible for GeoSPARQL inclusion. The profile should check for the information properties that are used in GeoSPARQL to place a 3D object in a correct way. If the profile is followed, translation from IFC to GeoSPARQL is guaranteed to work (assuming the entered values in the IFC are correct).

Such profiles are called "Informatieleveringsspecificaties" or "ILS" in The Netherlands.

## Actor

- Data Expert (who may not be a Geodata Expert)

## Preconditions

- I have an IFC file.
- I have some GeoSPARQL data.

## Postconditions

- I have integrated my IFC 3D data into my GeoSPARQL data. I was able to do this in a modern triple store, without having to write any code.

## Steps

1. I have an IFC file.
2. I use the GeoSPARQL profile for ILS to determine whether my IFC will automatically translate to GeoSPARQL 3D.
3. The profile indicated that my ILS lacks a geospatial reference.
4. I update my ILS to include a geospatial reference (or I ask the geospatial expert who gave me the ILS to change this for me).
5. When I use the GeoSPARQL profile for ILS again, I now pass all the profile checks. I can therefore add my ILS to a triple store that automatically stores it as GeoSPARQL 3D.
6. When I query my data, I can confirm that the ILS data now appears at the right spot, at the right height, in the right size, under the right angle, and with the correct textures applied.
7. I now upload a batch of 100 IFC files. 5 minutes later, I have a whole neighborhood of 3D buildings in my triple store, that others can query.

FransKnibbe:
Some notes:

1. IFC is a data exchange format for BIM.
2. ILS are text documents, but they can have machine-interpretable specifications: [IDS](https://technical.buildingsmart.org/projects/information-delivery-specification-ids/). An IDS could include specifications on how to make IFC data interoperable with geoSPARQL.
[...]

Issue #584, opened 2024-11-05
Triply Use Case 2: Support for textures


wouterbeek:
# Triply Use Case 2: Support for textures

## Description

In GeoSPARQL, I can represent and load 3D geometries in GML or WKT. But these geometries are boxes without texture. When I obtain textures, I cannot store them in relation to their corresponding 3D geometries with a standardized relationship. Originally reported by me in <https://github.com/opengeospatial/ogc-geosparql/issues/438>, including many insightful comments by others.

## Actor

- Data Expert

## Preconditions

- One or more 3D objects in a GeoSPARQL dataset.
- Textures that belong to these 3D objects, stored in separate files.
- Information about how the textures apply to the objects, e.g. where the texture should start and end (otherwise the front door would appear on the back of a building).

## Postconditions

- The 3D objects in my GeoSPARQL dataset have textures. Geospatial and texture data is fully integrated.

## Steps

1. I have 3D GeoSPARQL objects in my triple store.
2. I have textures for these objects in separate files.
3. I can integrate the textures into my GeoSPARQL dataset. Either by linking to the files, or by integrating the file content into a (binary) literal.
4. When I query my data, I can show the 3D objects on a map, together with their textures. The triple store knows how the textures should be applied to the 3D objects, as long as I have specified sufficient information for how this should be done.

ar-chad:
@wouterbeek you can describe your model in terms of CityGML standard and use Appearance module to include textures and texture coordinates, in particular https://docs.ogc.org/is/20-010/20-010.html#TexCoordList-section

There is ongoing work on CityRDF, which is going to turn this standard to an ontology https://github.com/ogcincubator/cityrdf

In short, CityGML adds additional semantics on top of GML. This allows you to specify if a particular geometry is a wall or a floor, etc. And the Appearance module, in particular allows you to add texture information to such descriptions.

ar-chad:
Here is an example of simple CityGML model with one Solid and appearance applied to it:

```
<?xml version="1.0" encoding="utf-8"?>
<CityModel
  xmlns="http://www.opengis.net/citygml/2.0"
  xmlns:gml="http://www.opengis.net/gml"
  xmlns:app="http://www.opengis.net/citygml/appearance/2.0"
  xmlns:dem="http://www.opengis.net/citygml/relief/2.0"
  xmlns:bldg="http://www.opengis.net/citygml/building/2.0"
[...]

Issue #583, opened 2024-11-05
Triply Use Case 1: Support for GLTF and other 3D formats


wouterbeek:
# Triply Use Case 1: Support for GLTF and other 3D formats

## Description

There is good support for 3D objects in BIM tools, and there is good support for geo data in GeoSPARQL triple stores. But there is not yet a good way in which I can combine the two. As a data expert, I am able to find files that contain 3D objects in standardized formats (e.g. GLTF), but I do not know how to load these into my GeoSPARQL triple store. When I look this up on the web, I end up in academic papers on how GLTF could be integrated with GeoSPARQL. There are multiple approaches, none of which is supported out-of-the-box by contemporary triple stores.

## Actor

- Data Expert

## Preconditions

- A 3D object in a file with a popular format (e.g. GLTF).
- A geospatial/GeoSPARQL dataset, in which I can meaningfully position my file.

## Postconditions

- A 3D object within my geospatial/GeoSPARQL dataset.

## Steps

1. I have obtained a file that contains a 3D object in a popular format (e.g. GLTF).
2. I can upload that file in a triple store, that may contain geospatial/GeoSPARQL data.
3. In a query, I can retrieve the 3D object, in combination with other geospatial/GeoSPARQL data (from my local store and/or from a remote endpoint).

situx:
Hi @wouterbeek and thanks for your use case!

We were thinking about supporting links to 3D models as resources for now.
You seem to propose uploading them to the triple store, which I think might slow down querying.
Do you mean e.g. serializing e.g. a PLY file and storing it as a String literal or do you mean to upload it somewhere accessible to the triple store but not inside the knowledge graph?

ar-chad:
@wouterbeek it would be great, if you could also try to add function signatures to your use cases.

FransKnibbe:
@situx: at Triply, we would like to be able to use 3D geometry serialisations as literals, in the same way as WKT, GML or GeoJSON literals. It would be good to have support for at least one 3D geometry serialisation that is a general format for 3D geometry. We have used glTF, which is a format that is also used in the 3D Tiles standard, so it is already something common in GIS. But PLY or OBJ could also be candidates. I guess formats that allow the most pure geometry encodings are best suited, to have an easy option to use such a serialisation purely for geometry (and exclude data that are needed for visualisation). I don't have a good insight in the various general 3D graphics formats, but OBJ seems well suited for compact geometry representation, and can also be used for point clouds.

[...]

Issue #571, opened 2024-10-18
Web3D and X3D Ontology


VladimirAlexiev:
Hi everyone! What do you think of this:

X3D Ontology
https://www.web3d.org/x3d/content/semantics/semantics.html
https://www.web3d.org/working-groups/x3d-semantic-web
https://www.web3d.org/x3d-semantic-web-public-assets
https://www.web3d.org/examples
https://www.web3d.org/x3d/content/examples/X3dResources.html#Examples
https://www.web3d.org/x3d/content/examples/X3dResources.html
https://www.web3d.org/x3d/content/examples/HelloWorld.ttl

These guys have done a ton of work that's relevant to GeoSPARQL 3D.
It seems more oriented towards VRML, human animation etc: https://www.web3d.org/working-groups. But certainly:
- CAD is in scope
- Geospatial is being actively developed https://www.web3d.org/working-groups/geospatial
- I saw a mention with some desire to address BIM

--

I found it in an old doc "Proposed extensions to GeoSPARQL 1.0":
https://github.com/opengeospatial/geosemantics-dwg/blob/master/white_paper/clause_4.adoc#extension-11-extending-the-geosparql-ontology-with-full-featured-3d-support

But it's not mentioned in https://github.com/opengeospatial/ogc-geosparql/issues/28 or https://github.com/opengeospatial/ogc-geosparql/issues/20.

cc @mathib @robmetalinkage @avillar @nicholascar @situx @peterbonsma @ar-chad @nataschake @RolfJonker

situx:
If you are referring to this document: https://github.com/opengeospatial/geosemantics-dwg/blob/master/CR585attachment.pdf
Yes, this is the document I wrote when we were founding the GeoSPARQL group. It was and is on my radar, mainly as a serialization format we might want to include.
Of course there might be much more to learn and if we can extract functions or vocabulary that is useful for GeoSPARQL, we can discuss it.

PR #611, opened 2025-01-25
3D Primitives Vocabulary and Alignments to IFC, CityRDF and X3D

situx:
This pull request starts to consolidate the following things:

1. A vocabulary simply called "Simple Features 3D" for 3D primitives common in other 3D vocabularies
   - Curve Types as extensions to the Simple Features Vocabulary (e.g., BSplineCurve)
   - Surface Types as defined from different curves (e.g. BSplineSurface)
   - 3D Primitives in use in other 3D formats: e.g., Cone, Cylinder, etc.
   - Literal types to encode common 3D types #583
   - Material and Texture classes, which may or may not be replaced by CityRDF Appearance Module see #584
2. Alignments to:
   - IFC also to collect concepts for the IFC Profile #590 
   - X3D #571 

Missing points:
- Are all geometry types included in this vocabulary?
- Did we cover all relevant literal types?
- Is the naming appropriate?
- Any additions to the alignments?


paulc-dstl:
The set of 3D shapes is looking good. But just to note, some 3D shapes (geometries) can be either solid or not. A cube for example may be a mesh rather than a solid object (ie the cube comprises six faces but is hollow). This is analogous to a polyline having coincident start and end points but not being closed. So I think we need some facets here to allow for polyhierarchy, bearing in mind also that some 3D shapes are neither meshes (ie have no faces) nor are solid, for example a 3D polyline or a pointcloud.

situx:
Thanks @paulc-dstl 
Then it would be better to classify them as 3DGeometries and use the property geo:isClosed and geof:isClosed to describe them?

situx:
Meeting 5th February:

- @ar-chad investigates whether it is possible to generalize the Appearance module of CityRDF for the use within GeoSPARQL and how to align it
- @jabhay investigates whether it is possible to extend the Simple Feature Access Vocabulary with 3D types as proposed in this pull request or whether a separate vocabulary namespace is needed
- @nicholascar  to follow up with Rob Atkinson on the review of the 3D primitives.
- @situx to separate the appearance module parts from the simple features 3D class hierarchy

ar-chad:
> Meeting 5th February:
> 
> * @ar-chad investigates whether it is possible to generalize the Appearance module of CityRDF for the use within GeoSPARQL and how to align it

I have started investigation on this https://github.com/ogcincubator/cityrdf/issues/15

nataschake:
Hi!
Here is a [separated module](https://github.com/ogcincubator/cityrdf/blob/master/citygml-owl/geosparql/geosparql-appearance.ttl) for Appearance ontology, taken from  [CityRDF](https://github.com/ogcincubator/cityrdf/issues/15#issue-2835936593) and defining all necessary classes within itself. CityRDF itself (mainly Core and Appearance) still needs to be carefully tested, but you can evaluate the idea.

ar-chad:
@nataschake GeoSPARQL is only one of the ontologies which would use this appearance ontology, CityRDF another one. There may be more in the future. We discussed making it independent of other ontologies so it could be imported by them or aligned with in another way.

situx:
Since the appearance module has now been excluded from this pull request I will try to finalize the 3D primitive vocabulary and then set this pull request as ready to review

PR #612, opened 2025-02-14
Issues 556 geometry extrusion

ar-chad:
Extrusion function without temporal dimension.


situx:
Hi @ar-chad 
thanks for the pull request.
However, it also includes parts that I prepared, but did not finish yet.
For example the spec changes that will need to be migrated.
It is fine to commit single functions, e.g. in functions.ttl and also as a section in the spec, but it would be better if you used a new branch sourced from geosparql-1.3 to state the pull request.

ar-chad:
> Hi @ar-chad thanks for the pull request. However, it also includes parts that I prepared, but did not finish yet. For example the spec changes that will need to be migrated. It is fine to commit single functions, e.g. in functions.ttl and also as a section in the spec, but it would be better if you used a new branch sourced from geosparql-1.3 to state the pull request.

OK, I'll change that to branching from geosparql-1.3. Thanks for letting me know.

PR #613, opened 2025-02-18
Added extrude and extrudeInTime functions.

ar-chad:


ar-chad:
Missing parts were result of some commit error. Now they are in the branch. I'll create new pull request with relevant parts of the spec as well.

PR #620, opened 2025-05-04
3D Literals for GeoSPARQL

### LINKED ISSUE #583: Triply Use Case 1: Support for GLTF and other 3D formats

### LINKED ISSUE #630: Registry needed for serialization extensions to allow for future extensibility of standard
ISO is concerned about allowing future serializations.

For each new serialization, the registry needs to include:
- Datatype
- Property
- Conversion Function

Q: Do we need to have existing serializations in the registry? Probably not. It would lead to redundancy and potential conflicts down the line.

Q: Should we have a separate serialization functions registry? Probably yes.

situx:
This pull request adds definitions of 3D literals in addition to the option to link to 3D files directly.
- PLY
- OBJ
- GLTF
- X3D

Closes #583

### DIFF
--- a/spec/sections/11-geometry_extension.adoc
+++ b/spec/sections/11-geometry_extension.adoc
@@ -38,6 +38,25 @@ requirement:: /req/geometry-extension/geocode-literal-srs
 requirement:: /req/geometry-extension/geocode-literal-empty
 requirement:: /req/geometry-extension/geometry-as-geocode-literal
 requirement:: /req/geometry-extension/asGeocode-function
+requirement:: /req/geometry-extension/ply-literal
+requirement:: /req/geometry-extension/ply-literal-default-srs
+requirement:: /req/geometry-extension/ply-axis-order
+requirement:: /req/geometry-extension/ply-literal-empty
+requirement:: /req/geometry-extension/geometry-as-ply-literal
+requirement:: /req/geometry-extension/asPLY-function
+requirement:: /req/geometry-extension/gltf-literal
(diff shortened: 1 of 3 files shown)


jabhay:
Jo to reach out to 3DIM people to get their input into recommended supported formats.

nicholascar:
These serializations need to be ported to datatypes in the datatypes register which has yet to be defined.

Issue #16, opened 2019-10-18
Extending GeoSPARQL by defining more vector literal types


lvdbrink:
**Official change request**: [OGC CR 585](http://ogc.standardstracker.org/show_request.cgi?id=585)

OGC Change Request Vector Literal GeoSPARQL

GeoSPARQL currently offers WKT and GML literal types which can be used to encode vector geometries.
However, geospatial data formats are very heterogeneous and a variety of other data formats deserve to be encodable in GeoSPARQL in our opinion.
In the following we list the most likely candidates:
- GeoJSON: Very common throughout the Web 
- GeoHash: Common has representations of vector data
- GPX: GPX Format used in GPS trackers
- KML Format by Google
- (H)(E)WKB/TWKB: Binary serializations of WKT often used as an internal storage format in databases

The following formats could be considered, but are in our opinion optional:
- LatLonText: Common format to display points in e.g. Wikidata or OSM 
- GeoURI: Defacto standard for mobile phone geo urls 
- Geobuf Format
- OSM Format: OSM XML
- Polyshape/EncodedPolyline: Format developed by Google to encode polylines/polyshapes
- SVG: Web standard for graphics in general 
- X3D: Standard to visualize 3D geometries

Implementations of most of the described literals can be seen in [an extension for rdf4j](https://github.com/i3mainz/rdf4j-postgis) and [an extension for Jena](https://github.com/i3mainz/postgis-jena). 

and in a proposed ontology [here](https://github.com/i3mainz/geosparql2.0)

Possibly, other literal implementations are useful and could be discussed.

The [attached PDF](https://github.com/opengeospatial/geosemantics-dwg/blob/master/CR585attachment.pdf) provides links to all the dataformat specifications and proposes how literal representations could look like.

jabhay:
MoSCoW poll created

[![](https://api.gh-polls.com/poll/01EJXAY5EFS4ACJ12MNZ8WM0ND/Must)](https://api.gh-polls.com/poll/01EJXAY5EFS4ACJ12MNZ8WM0ND/Must/vote)
[![](https://api.gh-polls.com/poll/01EJXAY5EFS4ACJ12MNZ8WM0ND/Should)](https://api.gh-polls.com/poll/01EJXAY5EFS4ACJ12MNZ8WM0ND/Should/vote)
[...]

PR #631, opened 2025-06-25
[EWKB CS Work Item] First Draft

jabhay:
This change is the first draft of the Work Item to adopt
EWKB as an OGC Community Standard. It relates to #16


situx:
- Mainz University is not an OGC member currently, so it would need to be removed in section 9
- We have discussed in the past that we might want to encode a URI like in the WKT literal for the EWKB literal and EWKB only defines an SRID in binary and not a URI of a CRS. Is that an issue?

jabhay:
Thinking about this more @situx , the reason we include an SRS URI in the asWKT string is because there is no way to embed an SRS into a legal WKT string. However, the EWKB does not suffer that same issue. I wonder whether consistency with our handling of asWKT is a good idea given capability to embed it in the binary string?

jabhay:
We have two options going forward to deal with the SRS URI problem.

1. Add the URI to the front like EWKT does.
2. Add a separate property to the ontology for SRS URIs, and leaving the EWKB untouched
3. Leaving the EWKB untouched and not having a SRS URI, depending instead on the SRID encoded within the EWKB.

Whilst 3 is suboptimal owing to potential lossiness for those SRS's that do not have an SRID, this pattern has been used for other encodings. We feel this is the most appropriate way forward for consistency.

@jabhay will raise the issue with Paul Ramsey so the EWKB authors and users are aware of the shortcoming, in case they have an interest in addressing it in future.

mperry455 (approved):
Looks good.

situx (approved):

PR #639, opened 2025-10-14
Add RCC*-9 relations to GeoSPARQL

situx:
Following the paper RCC*-9 paper https://www.mdpi.com/2220-9964/13/1/25 this pull request adds the vocabulary needed to express RCC9 relations.

Questions:
- Should we create RCC9 properties next to RCC8 properties for clarity even if the definition of the RCC9 property is equivalent to the RCC8 property? (I think so and I would link them as equivalentProperties)
- What are the relations of RCC9 to SimpleFeatures and DE9IM. We should express that in the vocabulary
- Do we keep the RCC8 relations in the GeoSPARQL 1.3 standard at all or should we simply provide RCC9?


situx:
Discussed 15th October 2025:
- Keep rcc8 properties and functions
- Add rcc9 properties and functions

nicholascar (changes requested):
I'm fine with the RCC9 work here in general - thanks @situx - but we have to do more regarding the vocab of functions. 

By just adding a Collection and the new Concepts, without adding them also to the Concept Scheme's hierarchy (top concepts + narrowers/broaders) we will confuse the vocab system (Rainbow) so we need to either:

1. create a RCC9 vocab, seperate to RCC* - with semantic relations between them
2. add RCC9 Concepts as top concepts to the RCC8 voc, as well as the Collection, and a Collection for just RCC8 terms, and rename the vocab.

I prefer the first option.

If there are any other thoughts, please reply to this or else I'll create the RCC9 vocab with relations.

nicholascar (approved):

PR #690, opened 2026-07-30
BFO alignment

### LINKED ISSUE #645: Definition of :Geometry (alignment with BFO)
Is `geosparql:Geometry` a region of space, or is it a set of coordinates that defines a region of space? 

I am attempting to reconcile GeoSPARQL with the [Comon Core Ontologies GeospatialOntology module](https://github.com/CommonCoreOntology/CommonCoreOntologies/blob/develop/src/cco-modules/GeospatialOntology.ttl). 
See the issue I have raised over there: https://github.com/CommonCoreOntology/CommonCoreOntologies/issues/758 

Common Core Ontologies are a set of 'mid-level ontologies' in the BFO/OBO family. 
They are significant because they are required to be used by parts of the US Government. 
The CCO GeospatialOntology has a mixture of concepts, some of which overlap with the scope of various OGC standards. 
It mentions `:asWKT` but this is pretty much isolated, and I'm trying to figure out how to relate GeoSPARQL more formally. 

A key distinction within the [class hierarchy of BFO](https://github.com/bfo-ontology/bfo-2020#basic-formal-ontology-hierarchy) is between 

- 'independent continuant' which has subclasses [spatial region](https://github.com/BFO-ontology/BFO-2020/blob/master/21838-2/owl/bfo-core.ttl#L721) & [continuant fiat boundary](https://github.com/BFO-ontology/BFO-2020/blob/master/21838-2/owl/bfo-core.ttl#L1062) (in BFO), and 
- 'generically dependent continuant` which has a subclass [information content entity](https://github.com/CommonCoreOntology/CommonCoreOntologies/blob/develop/src/cco-modules/InformationEntityOntology.ttl#L1213) (in CCO)

A generically dependent continuant exists in relation to some independent continuant(s), providing descriptive information about it (them). 

So the primary question is: can a `:Geometry` exist without thing(s) (e.g. `:Feature`) that it provides the geometry for? 
Related: may one `:Geometry` have multiple serializations (e.g. WKT _and_ DGGS)?

If yes, then - as @alanruttenberg [points out here](https://github.com/CommonCoreOntology/CommonCoreOntologies/issues/758#issuecomment-3554505798) - the current [textual definition of `:Geometry`](https://docs.ogc.org/is/22-047r1/22-047r1.html#_84414e8b-8ccc-407d-85b0-f3b474bba54d) is misleading, in particular as it refers to the direct positions that delimit the Geometry, rather than to the Geometry purely as a region of space. (He makes additional comments on the mention of the SRS and the adjective 'coherent'.) 

As you will see from my (perhaps overlong) comment that opens the [CCO GitHub issue](https://github.com/CommonCoreOntology/CommonCoreOntologies/issues/758), I had always assumed that `:Geometry` was the region of space, and the coordinates were merely a (perhaps imperfect) description of it. In BFO terms I had `:Geometry` aligned with 'spatial region', with the serialization being an 'information content entity' that describes it. If only one serialization is permitted, then clearly the coordinates **_are_** the geometry, and I'm maybe barking up the wrong tree.

dr-shorthair:
Closes #645

RDF is at https://github.com/dr-shorthair/geosemantics-semantic-resources/blob/main/resources/geosparql-swg/geosparql-1.1/source/alignments/bfo.ttl waiting for merge into https://github.com/opengeospatial/geosemantics-semantic-resources


dr-shorthair:
Wrong branch

PR #698, opened 2026-09-02
3dfunctions2

jabhay:


nicholascar:
I have merged in all recent changes so this PR is now essentially rebased onto the head of master, i.e. same original PR content with lots of other trivial changes like section IDs now added in. So it's ready to be reviewed for content.

